PR #13, opened 2016-08-19
set up rendering on os x

pkropf:
Here's a simple change set that allows for rendering on os x. It assumes that openscad and inkscape were both installed via the .dmg installation disk images from the respective websites.


scottbez1:
Thanks for helping make this cross-platform!

I'm not super familiar with how pull request builds work on Travis, but are you able to try re-running the build? I don't see why it should fail on your PR, and I haven't seen this kind of Inkscape error before, so wondering if it's just some kind of flake:

```
**
ERROR:verbs.cpp:797:static void Inkscape::FileVerb::perform(SPAction*, void*, void*): assertion failed: (parent != NULL)
Emergency save activated!
Emergency save completed. Inkscape will close now.
If you can reproduce this crash, please file a bug at www.inkscape.org
with a detailed description of the steps leading to the crash, so we can fix it.
Traceback (most recent call last):
  File "generate_2d.py", line 66, in <module>
    raster_svg,
  File "/usr/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['inkscape', '--verb=FitCanvasToDrawing', '--verb=FileSave', '--verb=FileClose', '--verb=FileQuit', '/home/travis/build/scottbez1/splitflap/3d/build/laser_parts/raster.svg']' returned non-zero exit status -11
```

3d/commands.py
```diff
@@ -0,0 +1,24 @@
+#   Copyright 2016 Peter Kropf
```
scottbez1: I think for consistency (and scalability) it might make sense to switch to a standardized header for all files in the repo, something like:
"Copyright 2015-2016 Scott Bezek and the splitflap contributors"?

I can go ahead and send a separate diff that will update the existing files, README, LICENSE, etc, to match, and that way it will be clear that the code in this repo is not from a single person. As far as per-file authorship goes, I think git history is probably a reasonable way to preserve attribution since it's automatic and avoids manual churn of header comments as new authors contribute/modify the code. 

I'm largely going off the discussion here http://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/ but open to alternative suggestions.

3d/commands.py
```diff
@@ -0,0 +1,24 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import platform
+
+
+if platform.system() == 'Darwin':
+    openscad = '/Applications/OpenSCAD.app/Contents/MacOS/OpenSCAD'
+    inkscape = '/Applications/Inkscape.app//Contents/Resources/bin/inkscape'
```
scottbez1: Is the double-slash needed here?

3d/generate_2d.py
```diff
@@ -45,25 +46,29 @@
     svg_output = renderer.render_svgs()
 
     # Export to png
-    raster_svg = os.path.join(laser_parts_directory, 'raster.svg')
-    raster_png = os.path.join(laser_parts_directory, 'raster.png')
+    cwd = os.getcwd()
```
scottbez1: Is specifying cwd necessary? I would have thought that in absence of an absolute path, any path would be treated as relative to cwd already?

I think there may be some existing issues with running this script from outside of the `3d` directory, but to resolve that kind of issue I think we'd want to make the paths relative to this file's location, e.g. make `laser_parts_directory` relative to this file, something like:
`laser_parts_directory = os.path.join(os.path.dirname(__file__), 'build', 'laser_parts')`.

That way even if you call the script from `/Users/foobar/` it will still output to the `3d/build/` folder within the repo (which is gitignored). Doing that would probably also require updating the `input_file` given to the `Renderer` above to properly reference the 'splitflap.scad' location as well.

Anyway, I think updating this script to work from outside the `3d` directory is worthwhile to do at some point (other scripts like `generate_stl.py` already support that), but probably a larger change, so for the purposes of this PR I'm wondering if this cwd specification is actually necessary, or if just the `commands` change when invoking inkscape/openscad is sufficient to make this work on OS X?

3d/commands.py
pkropf: I'm ok with changing the copyright.

3d/commands.py
pkropf: The double-slash is not needed, extra character left over from the cut and
paste.

3d/generate_2d.py
pkropf: There seems to be an issue with Inkscape on OS X whereby the working
directory for the subprocess isn't that of the launching program.  I'm not
sure why yet. I'm planning on also trying inkscape built through brew and
on Windows to see if there's any pattern to it.

scottbez1: Aha, that's really interesting! In that case this cwd usage seems find to me!

Another approach to try would be like I mentioned above with changing laser_parts_directory, but also throw in an `os.path.abspath`, so something like
`laser_parts_directory = os.path.abspath(os.path.join(os.path.dirname(__file__), 'build', 'laser_parts'))`.

Not a big deal though, I'm happy to merge as is.

scottbez1:
Thanks again for sending this. Did you want me to merge as-is, or were you planning to make adjustments?

pkropf:
Let me update the code with the changes we talked about and push a new
version to you. I'll see about getting that done in the next day or so.

- Peter

scottbez1:
Closing old PR

PR #22, opened 2018-05-03
v0.3 svg on Pokono P2 sized template

qwertyaardvark:
P2 template with v0.3 svg, fixed svg such that there are no double lines, ready for upload directly to Pokono


scottbez1:
Closing this as I'd like to avoid adding immutable release-related files to the repo itself; the Releases page is a better place for them, though I know that it doesn't currently have any Ponoko template files.

I do plan to include Ponoko template SVGs as part of the upcoming v1.0 release (in addition to the standard auto-generated plain SVG), though I'm waiting to hear back from Ponoko to double-check it's OK to redistribute derivative works of their template designs. Ideally I'll have the Travis build scripts auto-generate the Ponoko SVG file as well, to avoid needing to copy+paste the objects and align them properly, which will also mean Ponoko SVGs would automatically be available for download for the latest commit on every branch rather than just for Releases.

PR #40, opened 2018-09-28
fix(WebGL): Gracefully fail on missing WebGL

tknickman:
Amazing project @scottbez1! I stumbled across it when looking for a way to build my own splitflap (after seeing how crazy expensive the few mass produced ones are).

This is just a small PR fail gracefully and show an error when a user doesn't support WebGL (which apparently I don't in chrome, but I do on safari).

Here's a gif showing the two experiences (WebGL supported in safari on the left, WebGL not supported in chrome on the right)

gif demo

### DIFF
--- a/docs/js/viewer.js
+++ b/docs/js/viewer.js
@@ -7,143 +7,149 @@ var model_base_url = 'https://s3.amazonaws.com/splitflap-travis/branches/master/
 
 var debug = false;
 
+var viewer_canvas = document.getElementById("error-container");
 var canvas = document.getElementById("viewer_canvas");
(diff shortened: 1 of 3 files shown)


docs/js/viewer.js
```diff
@@ -7,143 +7,149 @@ var model_base_url = 'https://s3.amazonaws.com/splitflap-travis/branches/master/
 
 var debug = false;
 
+var viewer_canvas = document.getElementById("error-container");
```
scottbez1: nit: is this variable name accurate?

PR #44, opened 2018-10-28
Use PCB.GetLayerName to get layer name

seppestas:
Resolves "TODO: add the rest" by supporting all the layers available in the layout and removes the need for a custom layer name table.

Not that getting the layer name of a non-existing layer id will now result in the "BAD INDEX!" layer name iof. "Unknown(<layer_id>)". It might be a good idea to check for this or even to fail / raise an exception when attempting to plot a non-existing layer.

I found out about this function while working on https://github.com/productize/kicad-automation-scripts, a collection of useful KiCad scripts I started as a part of https://github.com/productize/kicad-file-render-service. Both these projects where based heavily on your work, thanks a lot for sharing it.

### DIFF
--- a/electronics/pcb_util.py
+++ b/electronics/pcb_util.py
@@ -26,23 +26,6 @@ from contextlib import contextmanager
 logging.basicConfig(level=logging.DEBUG)
 logger = logging.getLogger(__name__)
 
-_LAYER_NAME = {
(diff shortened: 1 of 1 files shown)


scottbez1:
Awesome, that's so much better, thanks!

PR #45, opened 2018-10-30
Wait until schematic plot exists with timeout iof waiting for 2 seconds

seppestas:
Typically plotting the schematic does not take that long so the 2 seconds wait makes plotting the schematic take way longer than it should.

Is should now also give a clearer error message when plotting the schematic failed since the script should not attempt to read a non-existing file when plotting times out.


scottbez1:
Hmmm, I'm seeing this in the failed travis build:

```
INFO:__main__:Select PDF plot format
INFO:__main__:Plot
INFO:__main__:Wait before shutdown
INFO:__main__:splitflap.pdf created
Shutting down....   
Done.
Written 197820 bytes
(197820 of which were video data and 0 audio data)
..
Goodbye!
DEBUG:export_util:Restoring /home/travis/build/scottbez1/splitflap/electronics/splitflap.sch
INFO:__main__:Rasterize
convert.im6: Postscript delegate failed `/home/travis/build/scottbez1/splitflap/electronics/build/splitflap.pdf': No such file or directory @ error/pdf.c/ReadPDFImage/677.
convert.im6: no images defined `/home/travis/build/scottbez1/splitflap/electronics/build/schematic.png' @ error/convert.c/ConvertImageCommand/3044.
```

which ultimately leads to a failed export:

```
subprocess.CalledProcessError: Command '['convert', '-density', '96', '/home/travis/build/scottbez1/splitflap/electronics/build/splitflap.pdf', '-background', 'white', '-alpha', 'remove', '/home/travis/build/scottbez1/splitflap/electronics/build/schematic.png']' returned non-zero exit status 1
```

If I'm understanding those logs correctly, it seems like KiCad creates the file right away, but maybe takes a while to write the real contents, meaning we try to invoke imagemagick before the file is a valid PDF? But actually, this kind of failure seems vaguely familiar to me. I think I may have tried a similar approach originally but gave up due to a similar issue and just went with a lame 2 second sleep instead? Wondering if you have any thoughts or ideas?

seppestas:
I was able to reproduce the issue of the pdf being created before eeschema is done with it.
I changed the commit so it uses [psutil](https://psutil.readthedocs.io/en/release-3.0.1/) to detect when eeschema closes the pdf plot.

electronics/scripts/export_schematic.py
```diff
@@ -72,13 +79,31 @@ def eeschema_plot_schematic(output_directory):
...
+def process_has_file_open(process, file_path):
+    for file in process.open_files():
+        if file.path == file_path:
+           return True
+    return False
+
+def wait_for_plot(eeschema_proc, plot_file, timeout=2):
+    begin = time.time()
+    ps = psutil.Process(eeschema_proc.pid)
+    while not process_has_file_open(ps, plot_file):
```
scottbez1: One thing I'm thinking of is what happens if kicad finishes quickly (so the opposite of what we initially saw)? Presumably if kicad was really fast it could open *and* close the file before we even call `wait_for_plot`, in which case it would time out thinking the file was never opened, right?

Not sure how realistic that concern is, but I'm wondering if it wouldn't be more robust to instead just have a single wait loop in this method, which will wait as long as *either* eeschema has the file open OR the output file is missing/0-bytes. What do you think of an approach like that? Then it would initially be looping waiting because the file doesn't exist, and then loop because kicad had the file open, and eventually exit when the file is closed and non-empty. Not sure if it's worth it so I'm curious what you think.

PR #53, opened 2018-12-04
Use webcolors module to support named colors in color exporter

gromgull:


3d/colored_stl_exporter.py
```diff
@@ -204,3 +216,7 @@ def mkdir_p(path):
             pass
         else:
             raise
+
+if __name__ == '__main__':
+    import sys
+    ColoredStlExporter(sys.argv[1], sys.argv[2]).run()
```
scottbez1: nit: argparse might be nicer for this, e.g. something like:
```
    parser = argparse.ArgumentParser(description='Export an STL file for each color in an OpenSCAD model')
    parser.add_argument('input_file', help='OpenSCAD file')
    parser.add_argument('output_directory', help='Output directory')
    args = parser.parse_args()
    ColoredStlExporter(args.input_file, args.output_directory).run()
```

3d/colored_stl_exporter.py
```diff
@@ -187,13 +192,20 @@ def get_transformed_file_path(original_path):
...
-        ]
+        if match:
+            return [
+                float(match.group('r')),
+                float(match.group('g')),
+                float(match.group('b')),
+            ]
+        if '"' in color and webcolors:
+            try:
+                c = webcolors.name_to_rgb(color[1:-1]) # skip the ""
```
scottbez1: nit: what do you think about using regex for this, so that it's more similar to the RGB_COLOR_REGEX usage above it, and then also can assert that the value looks something like `".*"` rather than just checking if a `"` exists anywhere in the string and then attempting to parse it as a webcolor assuming the first and last characters are `"`?

PR #61, opened 2019-01-29
add rotation offset option in splitflap module

henricazottes:
Hi,

While installing my flap screen, I had an issue with flaps being just about to flip but still stuck upward. Thus sometimes I had wrong letters displayed. Instead of fixing the front case itself I found more convenient to add an offset option when you declare a module. With this option I have been able to tweak the rotation so every flap do correctly flap when needed :)

```c++
// Last argument is the optional offset_steps value
SplitflapModule moduleA((uint8_t&)PORTD, 4, (uint8_t&)PINB, B00010000, 10); 
```

Remark: Consider testing it on your screen before merging, I cleaned the code and added a default value but I don't have my screen to test the final code (even if with the few modifications I think it's quite safe...).

### DIFF
--- a/arduino/splitflap/splitflap_module.h
+++ b/arduino/splitflap/splitflap_module.h
@@ -99,6 +99,7 @@ class SplitflapModule {
   uint8_t target_flap_index = 0;
 
   // Current position/destination. Numbers are modulo GEAR_RATIO_INPUT_STEPS
+  const uint32_t offset_steps;
   uint32_t current_step = 0;
   uint32_t delta_steps = 0;
 
@@ -131,7 +132,8 @@ class SplitflapModule {
     uint8_t &motor_out,
     const uint8_t motor_bitshift,
     uint8_t &sensor_in,
-    const uint8_t sensor_bitmask
+    const uint8_t sensor_bitmask,
+    const uint8_t offset_steps
   );
 
 #if HOME_CALIBRATION_ENABLED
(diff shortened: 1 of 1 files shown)


henricazottes:
@scottbez1 have you been able to have a look at it? The CI error concerns the electronics but I haven't touched it as you can see :/

scottbez1:
Sorry for not responding sooner.

I am going to close this since the recommended approach for adjusting positioning is to shift the sensor physically forward or back as needed, to change when the sensor is triggered relative to the spool position. The sensor adjustment should enable at least +/- 1 flap adjustment in very fine increments, and larger offsets can be fixed (albeit laboriously) by shifting flap positions within the spool.

This is preferable to a firmware solution because:
- the "offset" is stored physically as part of the module itself, so you can unplug that module and move it to a different position or swap modules without having to reprogram any offsets in code; modules are truly interchangeable this way
- offsets in code add complexity, and potential for confusion (e.g does a positive offset mean the position is shifted further towards the next flap, or towards the previous flap?)

henricazottes:
Better later than never :p

> The sensor adjustment should enable at least +/- 1 flap adjustment

For the records, my need back then was to move just a bit more than what the sensor would say. Because of tolerance in cards cut and frame cuts I would have cards stuck and it would display the wrong letter. But it wasn't for every letters, just some of them. By adding just a bit of rotation it was fixing this issue. But yeah I could have sanded the frame or re-cut the cards. In my very specific situation it was not as easy as changing the code.

I totally agree with your points, thanks for passing by and explain your vision :)

PR #70, opened 2019-06-14
ESP32 support

charafantah:
Add support for ESP32 boards and allow control over Serial Bluetooth

### DIFF
--- a/arduino/splitflap/spi_io_config.h
+++ b/arduino/splitflap/spi_io_config.h
@@ -138,7 +138,7 @@ inline void motor_sensor_io() {
   // using SPI to read byte-by-byte and shift the data accordingly.
   bool extra_bit = digitalRead(D6);
 #elif defined(ESP32)
-  // Same problem for the ESP32, same solution
+  // Same problem for the ESP32, same solution.
   bool extra_bit = digitalRead(MISO);
 #endif
   for (uint8_t i = 0; i < MOTOR_BUFFER_LENGTH; i++) {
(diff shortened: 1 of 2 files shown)


arduino/splitflap/spi_io_config.h
```diff
@@ -50,8 +50,17 @@
   #define IN_LATCH() {digitalWrite(IN_LATCH_PIN, LOW); digitalWrite(IN_LATCH_PIN, HIGH);}
 #endif
 
+#ifdef ESP32
+  #define NEOPIXEL_PIN (13) //Any pin will work
+  #define OUT_LATCH_PIN (12)//Any pin will work
+  #define IN_LATCH_PIN (27) //Any pin will work
+
+  #define OUT_LATCH() {digitalWrite(OUT_LATCH_PIN, HIGH); digitalWrite(OUT_LATCH_PIN, LOW);}
+  #define IN_LATCH() {digitalWrite(IN_LATCH_PIN, LOW);digitalWrite(IN_LATCH_PIN, LOW);digitalWrite(IN_LATCH_PIN, HIGH);}  //on ESP32, for some weird reason, it sometimes doesnt work if we do only 1 digitalWrite!
```
scottbez1: Hmmm, that sounds like potentially a timing issue, as I don't know why else an extra digitalWrite with the same value would be any different. Have you tried a delayMicroseconds in between the LOW and HIGH to make sure the latch pulse is long enough?

charafantah: Yes, it does look like a timing issue, or maybe I have a bad clone?
I did try the delayMicroseconds but it didn't work. I even tried to change pins because I thought there must be something else putting the pin high, but  still had the same problem.

I ordered more esp32 boards to try.

PR #72, opened 2019-08-03
Allow SENSOR_TEST without NEOPIXEL_DEBUGGING_ENABLED

ericyeargan:
Handy for verifying a spool's magnet orientation without the custom PCB.


scottbez1:
Thanks for this change, and sorry for taking so long to review it. I updated it with the latest master and made a few small tweaks before merging. Thanks!

ericyeargan:
No prob!

PR #88, opened 2020-09-10
Round motor window and CAD tweaks

dmadison:
Rounds the corners of the motor window (r=5) for slightly more stiffness and a nicer fit + finish. With a radius of 5 mm this should still be a loose fit around the motor (at least according to the model). On the flat panel the spool retaining wall was moved to the X center of the window to clear the rounded corners while still maintaining its shared bottom edge.

This also adjusts the lasercut panel so that the module connector pieces won't cut material out of the enclosure's left side. To do this the top and bottom pieces were moved down and the connectors were nested between the tabs in the top right corner. This keeps within the existing boundaries of the panel so it uses no extra material.

For what it's worth these new connector positions break down and start intersecting the other parts around a thickness of ~5, though the connector geometry itself seem to break down there as well.

Last but not least this removes the `center=true` arguments on `circle()` calls, which aren't needed and were creating warnings in OpenSCAD (2019.05).

Before:
sf-flat-tweaks-before

After:
sf-flat-tweaks-after

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -1026,6 +1026,7 @@ if (render_3d) {
 } else {
     panel_height = enclosure_length + kerf_width + enclosure_length_right + kerf_width + enclosure_width + kerf_width + spool_strut_width + kerf_width;
     projection_renderer(render_index=render_index, render_etch=render_etch, kerf_width=kerf_width, panel_height=panel_height, panel_horizontal=panel_horizontal, panel_vertical=panel_vertical) {
+        // Main enclosure (left, right, front)
         translate([0, 0])
             enclosure_left();
         translate([0, enclosure_length + kerf_width])
@@ -1034,6 +1035,7 @@ if (render_3d) {
             rotate([0, 0, -90])
                 enclosure_front();
 
+        // Top and bottom
         translate([enclosure_height + kerf_width + enclosure_length_right, enclosure_length + kerf_width + enclosure_length_right + kerf_width + 2])
             rotate([0, 0, 90])
             enclosure_top();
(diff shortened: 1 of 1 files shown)


3d/splitflap.scad
```diff
@@ -433,16 +434,23 @@ module enclosure_front() {
...
     }
+	
+	hull() {
+		x = -28byj48_chassis_radius - motor_hole_slop/2 + motor_window_radius;
+		y = [-28byj48_shaft_offset - motor_backpack_extent - motor_hole_slop/2 + motor_window_radius,
+			-28byj48_shaft_offset + 28byj48_chassis_radius + motor_hole_slop/2 - motor_window_radius];
+		
+		for (i = [0 : 3]) {
+			translate([x * (i%3 ? -1 : 1), y[i%2], 0]) circle(r=motor_window_radius, $fn=40);
+		}
```
scottbez1, line 452: nit: while this does avoid some repetition, it's only 1 line shorter and arguably harder to understand with 2 different modular arithmetic operations. I would personally probably lean toward just repeating the translate/circle statements 4 times for clarity:
```
        translate([x, y[0], 0]) circle(r=motor_window_radius, $fn=40);
        translate([-x, y[1], 0]) circle(r=motor_window_radius, $fn=40);
        translate([-x, y[0], 0]) circle(r=motor_window_radius, $fn=40);
        translate([x, y[1], 0]) circle(r=motor_window_radius, $fn=40);
```

scottbez1 (approved):
Really nice, thanks! I'm impressed with your ability to wade through this and make some non-trivial changes! One small nit inline; let me know if you disagree and I'm happy to merge as-is, otherwise I'll merge after an update.

3d/splitflap.scad
dmadison: You're absolutely right, writing out all four statements does make it more readable. I'll make that change now. :smiley:

PR #89, opened 2020-09-10
3D scripting with Python 3

dmadison:
Updates the '3d' scripts for forwards-compatibility with Python 3. Replaces 'print' with the function version from `__future__` and adds explicit `utf-8` encoding and decoding where required. While I was at it I also tweaked `colored_stl_exporter.py` to use `argparse` rather than `sys` for better usability from the command line, and added an `argparse` option to override the `thickness` parameter in `generate_2d.py`.

I also moved the '3d' Python scripts into the nested `scripts` folder and adjusted the scripts, Travis CI config, and README to use the new locations. This cleans up the SCAD directory and matches the hierarchy from `electronics`. This appears to pass the CI checks but let me know if anything is amiss.

I've tested these with both Python 2.7.18 and Python 3.8.0 on Windows (x64) and found no discrepancies. I also took a glance at the Python scripts in `electronics`, but it looks as if the latest stable version of KiCAD (5.1.6) still uses Python 2 for the scripting interface.


scottbez1 (approved):
:clap: :clap: :clap:

PR #92, opened 2020-09-15
Font Rework and SVG Generation

dmadison:
This PR overhauls how fonts are handled in the SCAD assembly, and adds a font generation feature for rendering SVG font files for cutting. The goal is to make it easy to generate custom split-flap fonts for desktop cutting machines (Cricut, Silhouette, etc.).

Changes:
* Added parametric font generation to SVG, for cutting your own letters (`generate_fonts.py` and `font_generator.scad`)
* Letter rendering is now handled by a single function, with an argument to trim and position for either top or bottom 
* Letter geometry / position now matches flap geometry, so both `flap()` and `flap_letter()` can be called from the same point without additional transformations
* Letters are now trimmed to flap outlines exactly, rather than using a 2D exclusion area for the opposite flap
* Fixed vertical text alignment and baseline position. Characters like ',' and '.' now sit at the baseline, and characters like 'Q' and 'j' can extend below it
* Letters are now parameterized (`flap_fonts.scad`):
    * Font selection
    * Letter scaling (X/Y)
    * Letter position offsets (X/Y from center)
* Letter position can now optionally compensate for the gap between flaps (`letter_gap_comp`)

Also included:
* In 3D mode, the spool assembly can now be hidden (`render_spool`)

With these new options I've tweaked the default letter scaling width (0.8 of previous) so that letters such as 'M' don't exceed the flap bounds. I've also nudged the X and Y offsets so the letters are better centered with the Roboto font.

Here's a comparison of the font versions in the main assembly (`render_flaps` at 1 for front flaps only).

### Main Assembly, Before (`master`, 008593c)
sf-font-before-008593c

All default settings.

### Main Assembly, After (`fonts`, eca0516)
sf-font-after-eca0516

`letter_gap_comp` set to `True`. Note the consistent slope of the '4' from top to bottom.

### Font Generation

`font_generator.scad` renders a string of text across an X/Y array of flaps. This uses the same flap and font settings as the main assembly and can be customized to change the font, scaling, position offsets, and more. Here is the assembly with 40 flaps (A-Z, 0-9, and punctuation) with Roboto:

splitflap-font-roboto

Default settings circa eca0516, with `letter_gap_comp` set to `True`.

And here's the same file after exporting to SVG (`python scripts/generate_fonts.py`. The letters are passed through the optimizer so the lines on the flap won't be cut twice:

splitflap-font-svg

The Python script allows overriding most of the parameters from the command line, so font generation can be automated if need be. The flap outlines are purposefully exported for alignment purposes.

---

I think this is fairly complete, but please let me know if you see anything amiss!


scottbez1:
This looks really exciting! I've been thinking about doing the letter gap compensation ever since I threw together the font rendering, but never got around to it since it was just used for rough visual checks; it's great to have this now that you're rendering flap designs!

Will try to review this weekend!

On a related note, have you thought about an option to render the flaps/font as 2 separate files for double-sided printing? I believe some people have done double-sided printing on a big sheet and then cut out the flaps from that (and I think this is what [Oat Foundry does for their flaps](https://online.fliphtml5.com/oxcvb/ymnn/files/large/3.jpg?1589212790)). I've also thought about potentially printing on sticker paper and applying a whole page at once to a bunch of die-cut flaps sitting in an alignment jig, which could also use the same double-sided rendering. The two changes needed would be to render the fonts with bleed off the edge of the flap, and to adjust the layout of the top and bottom half of each letter so they match up double-sided. Anyway, just figured I'd throw that out there!

dmadison:
>On a related note, have you thought about an option to render the flaps/font as 2 separate files for double-sided printing?

I hadn't thought about that! My goal with this was to generate the letters for cutting out of vinyl, but you're right you could absolutely modify it for double-sided printing.

I'm not familiar with how files are prepared for the double-sided printing process, but I think that the letter drawing function I wrote already places the top/bottom letters correctly. That is to say, both the top and bottom letters are drawn on the flap in the correct position without any sort of additional transformations. You'd only need to render half of the flaps, bump the letter 'half' value up by one in the function call, and then export the sheet again. It gets moderately more complicated if the whole sheet needs to be mirrored, but not by much.

For bleed, I suppose we could scale the 3D flap before performing the intersection? We'd need to do a bit of math against the flap dimensions if we wanted it to be in real units but it should be feasible!

3d/scripts/generate_fonts.py
```diff
@@ -0,0 +1,103 @@
...
+    parser.add_argument('-f', '--font', type=str, help='Name of font to use (no spaces)')
+    parser.add_argument('-t', '--text', type=str, help='String of text to generate')
+
+    parser.add_argument('-h', '--height', type=float, help='Font height, as scalar of total flap height')
+    parser.add_argument('-w', '--width', type=float, help='Font width, as scalar of total flap width')
+    parser.add_argument('-x', '--offset-x', type=float, help='Character offset from center, X-axis')
+    parser.add_argument('-y', '--offset-y', type=float, help='Character offset from center, Y-axis')
+
+    parser.add_argument('--ncolumns', type=int, help='Number of columns / characters per row')
+    parser.add_argument('--no-comp', action='store_false', default=True, help='Don\'t compensate for the gap between top and bottom flaps')
```
scottbez1, line 50: nit: it's slightly confusing that the flag is called "no comp" but defaults to True and stores False when specified, and is then used as `'letter_gap_comp' : args.no_comp,`. It functions correctly, but looks like a bug since "comp" is being set to the value of "no comp". Can you make it `store_true`, default to `False`, and use it as `letter_gap_comp: not args.no_comp`? Alternatively, I think you could set `dest` to "comp" so the value would be referenced as "args.comp" instead of "args.no_comp", but that still seems slightly confusing to me.

3d/splitflap.scad
```diff
@@ -100,6 +100,7 @@ num_flaps = 40;
 
 flap_hole_radius = (flap_pin_width + 1) / 2;
 flap_gap = 1;
+function get_flap_gap() = flap_gap;
```
scottbez1, line 103: Ah interesting; is this a workaround to expose the flap_gap variable when `use`ing splitflap.scad rather than `include`ing it? If so, mind adding a comment to that effect here for clarity?

3d/splitflap.scad
```diff
@@ -361,10 +362,32 @@ module flap() {
...
+                translate([flap_width/2, -flap_pin_width/2, 0]) {
+                    rotation = (half == 2) ? -180 : 0;  // flip upside-down for bottom
+                    gap_comp = (letter_gap_comp == true) ? -flap_gap : 0;
+                    translate([0, gap_comp, 0])
+                        rotate([0,0,rotation])
+                            draw_letter(letter);
+                }
+            }
+        }
+        else {
```
scottbez1, line 388: style nit
```suggestion
        } else {
```

3d/splitflap.scad
```diff
@@ -361,10 +362,32 @@ module flap() {
...
+    linear_extrude(height=0.1, center=true)
+        translate([0, -flap_height * letter_height, 0])  // valign compensation
+            scale([letter_width, 1, 1])
+                translate([letter_offset_x, letter_offset_y])
+                    text(text=letter, size=flap_height * letter_height * 2, font=letter_font, halign="center");
+}
+
+module flap_letter(letter, half = 0) {
+    translate([0, 0, flap_thickness/2 + eps]) {
+        if(half != 0) {  // trimming to top (1) or bottom (2)
```
scottbez1, line 376: style nit
```suggestion
        if (half != 0) {  // trimming to top (1) or bottom (2) 
```

3d/splitflap.scad
```diff
@@ -922,64 +911,61 @@ module split_flap_3d(letter, include_connector) {
...
-                    }
-                    if (i == 0) {
-                        color([0,0,0]) {
-                            translate([0, flap_pitch_radius + flap_thickness/2, 0]) {
-                                letter_top_half();
+                        rotate([360/num_flaps * i, 0, 0]) {
+                            translate([flap_width, flap_pitch_radius, 0]) {
+                                rotate([flap_rendered_angle, 0, 180]) {
+                                    flap();
+                                    if(i == 0) { flap_letter(letter, 1); }  // 1 = top
```
scottbez1, line 918: style nit
```suggestion
                                    if (i == 0) {
                                        flap_letter(letter, 1);  // 1 = top
                                    }
```

3d/splitflap.scad
```diff
@@ -922,64 +911,61 @@ module split_flap_3d(letter, include_connector) {
...
                             rotate([-90, 0, 0]) {
                                 flap();
-                            }
-                        }
-                        if (i == 1) {
-                            color([0,0,0]) {
-                                translate([0, flap_thickness/2, 0]) {
-                                    letter_bottom_half();
-                                }
+                                if(i == 1) { flap_letter(letter, 2); }  // 2 = bottom
```
scottbez1, line 932: style nit:
```suggestion
                                if (i == 1) {
                                    flap_letter(letter, 2);  // 2 = bottom
                                }
```

3d/splitflap.scad
```diff
@@ -361,10 +362,32 @@ module flap() {
...
+}
+
+module flap_letter(letter, half = 0) {
+    translate([0, 0, flap_thickness/2 + eps]) {
+        if(half != 0) {  // trimming to top (1) or bottom (2)
+            intersection() {
+                flap();  // limit to bounds of flap
+                translate([flap_width/2, -flap_pin_width/2, 0]) {
+                    rotation = (half == 2) ? -180 : 0;  // flip upside-down for bottom
+                    gap_comp = (letter_gap_comp == true) ? -flap_gap : 0;
```
scottbez1, line 381: I'm sure I'm missing something, because the renderings look good, but why is the full flap_gap offset applied to both the top and bottom half? My naive intuition is that each would get a `flap_gap/2` compensation to make up for the total `flap_gap` loss?

scottbez1 (approved):
Looks like a great addition! A few small nits inline.

scottbez1:
Regarding bleed, I didn't realize when reviewing that the letter rendering is using a 3d intersection rather than 2d; I think it'd be preferable to do the intersection operation in 2d before extruding the letter/flap, which would then also mean you could apply the 2d `offset` operator to the flap shape before intersecting to specify a bleed distance in real units. That would also more closely match the guidance from the README of preferring 2d operations when possible and using an extrude as "late" as possible (`In general, solid objects such as the enclosure sides or spool components are built from 2d primitives and then extruded to the appropriate thickness for 3d rendering, rather than using 3d primitives.`)

3d/scripts/generate_fonts.py
dmadison: Agreed, changed.

3d/splitflap.scad
dmadison: Yes, exactly. Resolved with b524897.

3d/splitflap.scad
dmadison: Good catch. This was a bad assumption on my part - I assumed `flap_gap` was naively the gap between the flaps edges and for whatever reason you chose to store it halved. It actually refers to the gap between the edges of the flap pin _holes_ on the spool, which would otherwise be tangent.

I've refactored that variable as `flap_hole_separation` in both the main file and in `spool.scad` (6bc0e22), and created `flap_gap` as a new variable that represents the gap between the flap edges for the visible flap pair (44b1698). This new value is then halved in the letter function as expected.

dmadison:
You're right, it would be better and faster with a 2D intersection. I'll have to split up the `flap()` function but that shouldn't be an issue.

scottbez1:
Thanks!

PR #93, opened 2020-09-18
Thickness Fixes (+ 3.0 default)

### LINKED ISSUE #76: Using 3.0mm MDF (Ponoko no longer carries 3.2mm)
The OpenSCAD script and ordering guide specify 3.2mm MDF for the laster-cut parts. Unfortunately, it looks like Ponoko no longer carries 3.2mm. In some commentary in the ordering guide @scottbez1 suggests that it's possible that the material was always 3.0mm and that the Ponoko was just listing it incorrectly as 3.2mm.

So, I tried ordering a parts using 3.0mm MDF with the latest release but things don't fit quite right - there are visible gaps in the slots and a good amount of give when you lightly twist the spindle. (FYI, when measuring the material thickness with calipers, it's right around 2.99mm).

I've also tried modifying the OpenSCAD to change the material width to 3.0mm and update the kerf. I ordered parts with that updated pattern and things fit better although there's very little/no clearance between the spindle and sides. I had to reduce the top spindle speed to keep the motor from stalling - I suspect because the spindle is rubbing slightly against the sides. You can view the commit with the OpenSCAD change [here](https://github.com/ericyeargan/splitflap/commit/0d5985a39a45576651774b16db48451103f24ab1).

Curious if anyone else has dealt with this. My ultimate goal is to build 12 modules. Right now, I'm trying to decide wether to mess with the 3.0mm SCAD to increase the spindle clearance or to just live with a loose fit on the 3.2mm SCAD (maybe through some liberal CA glue application).

P.S. Despite these issues, I've been having a blast with this project. Thanks @scottbez1 and all of the contributors!

dmadison:
This PR fixes a few issues relating to panel thickness:

* Spool width now accounts for flap width. If the calculated spool width is smaller than the flap width, the enclosure width and front opening do not shrink further (ceb3664). This also comments out the `legacyAssert` for spool width which is no longer needed.
* Flap notch height and depth are now static and reflect the die-cut flap dimensions (773f4df). The flap height can be set back to auto with the `flap_notch_height_auto` flag.
* Enclosure side bolts are now positioned properly for all thicknesses (e87928c)
* The enclosure vertical inset (distance between top/bottom of front and top/bottom pieces) now has a minimum value to prevent the bolt heads and nut edges from extending past the bounds of the enclosure at thinner thicknesses (ee7e685)
* Sets the default thickness to 3.0, matching Ponoko's current MDF availability (8172e21). Tentatively closes #76

### Comparison, Standard Thickness ([master](https://user-images.githubusercontent.com/24282108/93540427-a1158300-f921-11ea-92c2-aa84296c6a84.png) / [8172e21](https://user-images.githubusercontent.com/24282108/93540448-aecb0880-f921-11ea-87ea-e01e4a5b29f0.png)):

sf-thickness-comp

(Take the tab edges of the flaps with a grain of salt, they're small features that render unevenly depending on the camera position)

### Comparison, 1.0 mm Front ([master](https://user-images.githubusercontent.com/24282108/93541001-56950600-f923-11ea-9785-6334ceec92b7.png) / [8172e21](https://user-images.githubusercontent.com/24282108/93540995-5432ac00-f923-11ea-97e8-16b52a0c9b35.png)):

sf-thickness-1-mm-comp

### Comparison, 1.0 mm Rear ([master](https://user-images.githubusercontent.com/24282108/93541078-7e846980-f923-11ea-898c-86d136d0cbbf.png) / [8172e21](https://user-images.githubusercontent.com/24282108/93541083-80e6c380-f923-11ea-9f62-fdd6cd5a03ab.png))

sf-thickness-1-mm-rear-comp


scottbez1:
Everything in the summary sounds and looks great! By the way, there's some discussion around changing this value in slack if you haven't already seen that: https://splitflap.slack.com/archives/CEMSQUP1C/p1590126017050000

It's a slightly breaking change, which I'm open to making, but will need to make sure it's clearly messaged to make sure anyone who has built previous v0.5 or v0.6 modules knows that the latest design will not fit compatibly with those older versions.

Will try to review this over the weekend.

dmadison:
Unfortunately I don't have slack so I haven't seen the previous discussion :grimacing:

My view is that most people looking to build their own would be starting fresh and not adding to an existing set of displays, in which case the model should reflect current materials and availability. Plus this only breaks backwards-compatibility so long as you're using the release files and not generating them yourself -- if you set the thickness back to 3.2 (with that slop-adjustment commit above!), everything should be compatible with the older releases.

scottbez1:
Here's the slack invite link (Slack doesn't support fully public groups for some reason, so the next best thing is to publicly share the invite link :shrug:) https://join.slack.com/t/splitflap/shared_invite/zt-dpvol87b-3zUaxXrUd8WauPXr1uBj5Q

dmadison:
I just read through the Slack discussion. Are you saying that the material was always 3.0 mm? I wonder how much of this comes down to the thickness tolerance of the material. Their [current material page](https://www.ponoko.com/materials/mdf-fiberboard) says the thickness varies between 0.11" (2.8 mm) and 0.13" (3.3 mm).

We could do a half-change where the panel thickness is updated but all of the derived values / clearances use a value that's 0.2 mm larger, but that's quite messy.

Personally I'd err on the side of accuracy and try to have the model reflect the actual design as accurately as is reasonable. If it's never updated future changes that rely on the thickness value would either have to ignore or compensate for that 0.2 mm difference on their own, which is just asking for trouble.

scottbez1:
Yeah, as far as I can tell the material has always been 3mm and their website was wrong before.

I agree that it's best to update the dimensions to match. I think it's just a matter of making sure the change in dimensions is noted clearly in the ordering instructions when a new release is created with this change.

dmadison:
In that case I think we need a different solution. As-is the assembly has zero clearance between many of the joints. If the assembly was rendered with a thickness value of 3.2 and ordered with a 3.0 thick panel, that gives you 0.2 mm of clearance for each joint. That's around 8 mil in imperial, which is in the ballpark of what you want for tolerance in an assembly anyways. If you trust the accuracy of your calipers you can verify that on your own displays (plus or minus the tolerance of the kerf value). I've pushed a commit to a ['thickness-real'](https://github.com/dmadison/splitflap/tree/thickness-real) branch, based off of `v5.0`, which changes the panel thickness extrusions to 3.0 while keeping all of the other thickness values at 3.2. This should give a rough look at how the model differs from the assembly as-built.

I think the solution is to lower thickness to 3.0 and modify most `thickness` instances to add clearance, probably in the area of 0.1 to 0.15 mm (4 mil to 6 mil). That way the model's panel thickness is accurate and the enclosure has predictable fitment tolerances.

For the time-being however I think the "3.0 as default" change should be reverted until those calculations can be updated. This PR can then be merged as-is to fix these other issues, as the 2D output at 3.2 should be identical with master.

scottbez1 (approved):
Everything here looks great. We can continue to discuss the plan forward before merging though.

scottbez1:
I'm actually inclined to go with 3mm as the default. Most functional dimensions are between two cuts, so material thickness shouldn't play a factor in those, so I'm not super concerned about fitment issues from changing 3.2mm to 3mm. Material thickness used to play a critical role in the spool spacing, but that was mostly fixed in https://github.com/scottbez1/splitflap/commit/325f6404056e4ba2e5cdf7afe7849fca1c922211#commitcomment-39978942 (though it's true that material that is oversized relative to the design dimension will still cause issues, which is more likely to happen when the design thickness is reduced to 3mm...)

I think the main place where it would come into play now is in the joints, but I think those should be just about right when reduced to 3mm, considering there has historically been a gap:
IMG_20200926_140506
IMG_20200926_140824

The default kerf correction for Ponoko MDF is maybe slightly too high (I measured ~3.19mm rather than 3.2 on some of my old Ponoko parts -- see below), which could cause fitment issues if design thickness is reduced to 3mm without fixing kerf; on that topic though, I'm thinking adding very small tolerances into the model may be a bit optimistic given the variability of laser cuts (even in different locations on the same piece/machine things like perpendicularity issues with the laser beam/mirrors can cause issues), and especially for soft MDF. Acrylic is a slightly different story since it's brittle, but to me it seems like it may be simpler to just handle fitment tolerance via the kerf correction factor (which is pretty important to get right anyway for different materials or laser-cutters) rather than adding the complexity of adding joint tolerances everywhere in the model. It's not technically "correct," but given the materials involved and the precision of measurement and cutting that most people will be using for this project, it feels like a reasonable tradeoff to make in the name of simplicity to me, but I'm not a mechanical engineer.

IMG_20200926_140632

Thoughts?

dmadison:
I agree that the joints are the crucial parts for the thickness change, we just need to figure out how best to handle them. (The spool geometry is as well, but that's handled by another commit in this PR.)

That vertical gap is definitely because of the thickness difference. I'd be curious to see the comparison between the horizontal opening on those tabs vs the horizontal bosses on the top pieces. How much clearance is there as manufactured?

>I'm thinking adding very small tolerances into the model may be a bit optimistic given the variability of laser cuts (even in different locations on the same piece/machine things like perpendicularity issues with the laser beam/mirrors can cause issues), and especially for soft MDF.

Think of it the other way around. You add in small allowances like that *because* the machinery can be imprecise. By bumping the clearance out a bit the resulting cut can be either tighter (close fit) or looser (free fit) but still accommodate the part. If you *aim* for a 0 fit, the close ('-') tolerance on that cut line will be tighter than the minimum thickness of the joint.

Let me put it another way. With the current no-clearance fitment, assuming that both thickness and laser kerf as-cut are accurately represented in the model, the resulting pieces will not fit together (or would be press-fit in the case of the MDF). Those joints only work because the values are inaccurate - the real-life openings are larger than they are in the model.

I'm not a big fan of adjusting this with the kerf setting because that reduces precision and might have repercussions elsewhere in the design. I think a few small offsets for each joint, on-center, is all that's needed.

The spool might be a bit more complicated as there are no bolts holding it together, but the general idea is the same.

dmadison:
I've pushed commits to add clearance for all of the mating enclosure pieces: the front and side tabs, the spool strut tabs, the spool joints, and the connectors. These are all 0.1 mm (0.004"), which is probably a good number for the screw-retained joints but may be a tad too high for the spool. I've also updated the kerf value with the latest from Ponoko, which is a flat '0.2' for both MDF and acrylic (all thicknesses).

Do you recall the reason for the '0.02' in the `kerf_width` calculation?

dmadison:
I've ordered a copy of the enclosure in acrylic from Ponoko based off of commit 63fcb01 - 3 mm thickness, 0.18 mm kerf, 0.1 mm clearance on all mating surfaces. (Should have done 0.2 mm kerf on the nose but I forgot to remove the `- 0.02 mm` bit.) Expected to arrive in a little over 2 weeks.

I'm going to hazard a guess that the spool connections at the least are going to be a little loose, but it'll be good to get a manufactured enclosure in hand for comparative measurements.

dmadison:
There was an issue with the cutting on my first enclosure, but I'm now back and with enclosure in hand! Ordered from Ponoko out of 3 mm matte black acrylic with a kerf value of 0.18 (0.2 - 0.02, current on branch).

Ponoko gives their kerf spec as 0.2 mm for 3 mm acrylic, although on cut parts I received features that are on average offset by 0.10 mm from the SVG, which makes cutouts undersized by about -0.08 mm (again, on average). With the additional clearance in the design *most* features work as designed, with three exceptions:

1. The magnet hole. This is intentionally undersized, but is possibly undersized too far. I don't have a set of gauge pins, but my calipers ballparked it at 3.75 mm (0.25 mm / 0.010" undersized). Pressing the magnet into place was difficult and caused the spool to crack internally.

2. The motor hole. Also intentionally undersized (zero fit) although less so than the magnet. Attempting to fit my spool on the motor created a fracture through the rest of the spool, rendering it unusable.

3. Less of an issue, but the front face is a slight press-fit rather than a free-fit.

All three of these issues will, hopefully, be resolved with a more accurate kerf setting. Although that's tricky to nail down from an online supplier when each order takes several weeks to get back. Some of this discrepancy is likely explained by the angle of the kerf, as only the widest edges are measured by the calipers (and, for that matter, are significant for fitment from the perspective of the edge).

It's also probable that all three of these issues won't seriously affect the design if it's cut out of a less brittle material such as MDF. Although it would still be better to fabricate the design as accurately as possible.

For the time-being I'd recommend dropping the kerf adjustment down, probably to 0.1 or 0.12. Ponoko's MDF may require a different value.

---

I've mentioned this before, but the material thickness tolerance is worth noting. For the matte black acrylic for example, Ponoko says it can [vary between 2.2 and 3.5 mm](https://www.ponoko.com/materials/black-matte-acrylic). For the two enclosure sets I received, one was around 2.95 - 3 mm while the other is around 2.75 mm. I also ordered some other 3 mm acrylic samples for color comparison, and those vary as well between 2.45 mm (blue) and 3.01 mm (red), with most of them averaging around 2.7 mm.

3d/splitflap.scad
```diff
@@ -248,7 +254,7 @@ module m4_captive_nut(bolt_length=m4_bolt_length) {
 
 // ##### Struts for bracing spool #####
 module spool_strut_tab_hole(narrow) {
-    square([thickness, narrow ? spool_strut_tab_width_narrow : spool_strut_tab_width], center=true);
+    square([thickness + spool_tab_clearance, narrow ? spool_strut_tab_width_narrow + spool_tab_clearance : spool_strut_tab_width + spool_tab_clearance], center=true);
```
scottbez1, line 257: Hmmm, I wonder if it would make sense to define the clearance values not on a per-part basis, but as 2 high level parameters, something like:
- thickness_clearance
- cut_clearance

Since cuts should be consistent and much lower variability than thickness, it seems like it would make sense to use a smaller clearance value for cut-to-cut joints (where dimensions are entirely controlled by cuts) compared to cut-to-thickness joints (where dimensions rely on inherent material thickness and therefore need a larger tolerance to ensure fit).

This part of the model involves both, so it would change to something like:
```suggestion
    square([thickness + thickness_clearance, narrow ? spool_strut_tab_width_narrow + cut_clearance : spool_strut_tab_width + cut_clearance], center=true);
```

scottbez1 (approved):
This is great, thanks so much! I am going to merge it as-is now, but let's continue discussing the thickness variability that you mentioned in your last comment and I referenced in my inline comment here.

3d/splitflap.scad
dmadison: Maybe, but the variability for both is wide enough that I'm not sure making the distinction would be beneficial. Both the kerf and the thickness are precise but not accurate, so assuming that the initial values are correct the tolerances should be roughly the same for both. The bigger overarching problem is that the design relies on friction-fit parts which necessitate accurate manufacturing.

Issue #58, opened 2018-12-20
Kicad 5


scottbez1:
The electronics projects are currently Kicad 4, but Kicad 5 is the latest version.

It would be nice to update all of the files to Kicad 5 and then verify that the schematics, footprints, and pcbs, along with all the Travis CI integration, still work correctly.

whitebird:
I've opened up the sensor PCB in KiCad 5. When opening a 3D view it throws the following error;
<img width="532" alt="screenshot 2019-02-04 at 14 57 51" src="https://user-images.githubusercontent.com/1058174/52212677-62869e00-288d-11e9-915f-496826df5cc3.png">

This is probably due to the yellow lines from the Edge.cuts layer not being a closed square. Can this be for mass producing the piece? I don't have KiCad 4 to compare the issue.

<img width="740" alt="screenshot 2019-02-04 at 14 58 15" src="https://user-images.githubusercontent.com/1058174/52212687-66b2bb80-288d-11e9-8cb2-4418b5f837e7.png">

scottbez1:
Yeah, those weird edge lines were from early experimentation with panelizing the board. IIRC the current panelization script ignores the original board outline layer entirely and redraws new lines based on the panelization parameters, but I'm not at my computer to double check at the moment. Regardless of the current behavior of the panelization script, the right thing to do is to fix that board outline, so I'll make a note to take care of that at some point.

scottbez1:
Closing in favor of #138

PR #95, opened 2020-09-20
Motor flip

dmadison:
This PR flips the motor upside-down while maintaining the same shaft axis. This has two benefits: it directs the motor wires downwards for easier cable management, and it slightly reduces the assembly's center of gravity.

To accommodate the motor's new position I also moved the sensor PCB and spool magnet to align with the top of the spool, with the PCB's connector facing upwards.

### Comparison ([before](https://user-images.githubusercontent.com/24282108/93691670-7f99d000-fab6-11ea-9d64-a391527e8861.png) / [after](https://user-images.githubusercontent.com/24282108/93691675-92140980-fab6-11ea-94ba-99c0b7047b02.png)):

sf-motor-flip-comp

---

I'm not entirely sold on this change, mostly because of how it affects the sensor PCB. Although it's nice to have the motor wires exit downwards, it moves the sensor PCB and its connector higher than the motor wires were originally, which somewhat defeats the point...

For a stopgap the right-angle pins could be replaced by the column from a multi-row right angle pin header ([e.g.](https://www.digikey.com/product-detail/en/samtec-inc/TSW-104-08-T-T-RA/SAM13035-ND/6691863)), although chopping up a part rather than using something off the shelf is less than ideal. Another option would be to solder the wires directly to the board instead of using a connector, which aids in the wire management but makes the build less modular.

Otherwise the PCB would need to modified to reorient the connector, probably by making the whole thing longer and narrower and having the connector exit out the back.

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -316,7 +316,7 @@ module flap_spool_complete(captive_nut=false, motor_shaft=false, magnet_hole=fal
             }
             if (magnet_hole) {
                 // Hole for press fit magnet
-                translate([magnet_hole_offset, 0]) {
+                translate([-magnet_hole_offset, 0]) {
                     circle(r=magnet_hole_radius, $fn=15);
                 }
             }
@@ -521,9 +521,9 @@ module enclosure_left() {
 
 
             // PCB mounting holes
-            translate([enclosure_height_lower - magnet_hole_offset - pcb_hole_to_sensor_y, enclosure_length - front_forward_offset - pcb_hole_to_sensor_x]) {
+            translate([enclosure_height_lower + magnet_hole_offset + pcb_hole_to_sensor_y, enclosure_length - front_forward_offset + pcb_hole_to_sensor_x]) {
                 rotate([180, 0, 0]) {
-                    rotate([0, 0, -90]) {
+                    rotate([0, 0, 90]) {
                         pcb_cutouts();
                     }
                 }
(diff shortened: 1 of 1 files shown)


scottbez1:
Hmmm, yeah, agreed that flipping the sensor PCB isn't ideal. I think something "creative" could be done to keep the existing PCB but rotate it so the wires come out the back with the slots adjusted accordingly:
pcb_rotated

but that's pretty gross and not something I'd want to keep long-term, and it might not even fit with the motor there...

Modifying the PCB is certainly an option, though I worry about confusion when people order parts if there are multiple similar variants. Might be possible to just keep the existing PCB layout and add additional connector holes in a different location so it's backward compatible? That would certainly be my preference, though any PCB modification really ought to start with migrating all of the kicad projects from version 4 to 5 (which isn't hard by itself, but some of the CI tooling also needs to be adapted/replaced. I've already started some of that tooling work in other branches where I'm working on the newer ESP32 controller PCB, including adding a new 3d PCB rendering export script, but the biggest problem I hit was the 3rd-party panelize.py script has not been updated for the kicad 5 file format, so that probably needs to be completely replaced by [KiKit](https://github.com/yaqwsx/KiKit))

dmadison:
>Might be possible to just keep the existing PCB layout and add additional connector holes in a different location so it's backward compatible?

Possibly, although it's going to be a tight fit. Quick test in KiCAD:

sf-sensor-pcb-pivot-test

The courtyards *just* clear each-other, so it's technically do-able but they're pretty packed in there. You could use an SMD connector on the opposite side of the board but then you lose the anti-rotation from the pins out of the back.

For a modified PCB I was thinking you could either [put the header inline with the bolt](https://user-images.githubusercontent.com/24282108/93949101-1613fe80-fd0e-11ea-8f48-fa95654425d7.png) and then 'step' the cutout in the enclosure, or [angle the PCB header so the pins are the same width as the bolt.](https://user-images.githubusercontent.com/24282108/93949099-144a3b00-fd0e-11ea-9896-8588bfb9949f.png) You could extend that PCB higher to keep the original header, though the silkscreen and connections will have to be mirrored depending on which 'version' of the enclosure is being used. It also makes the PCB more expensive.

dmadison:
Closing this for now. I don't think this is a viable change with the current sensor PCB design, and I don't think it's worth investing time to redesign the sensor PCB without upgrading the repo to use KiCAD 5 (#58) (or even 6, which is due out early next year).

PR #100, opened 2020-11-14
Mirror enclosure sides for lasercutting

dmadison:
This PR mirrors the left, right, and front enclosure pieces on the lasercut layout so that all outward faces of the design are facing 'up' when cut. This is useful when cutting asymmetric materials like Ponoko's matte black acrylic. 

#### Before:
sf-flat-pre-mirror

#### After:
sf-flat-post-mirror

The only change to the design itself is that the spool strut cut out of the right enclosure, which has not been mirrored along with the right side. This keeps it at the same orientation as the spool struts at the top, although it effectively mirrors it around the part's Y axis. This shouldn't impact the design's functionality.

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -1022,17 +1022,16 @@ if (render_3d) {
     panel_height = enclosure_length + kerf_width + enclosure_length_right + kerf_width + enclosure_width + kerf_width + spool_strut_width + kerf_width;
     projection_renderer(render_index=render_index, render_etch=render_etch, kerf_width=kerf_width, panel_height=panel_height, panel_horizontal=panel_horizontal, panel_vertical=panel_vertical) {
         // Main enclosure (left, right, front)
-        mirror([1, 0, 0]) {
+        mirror([1, 0, 0])
             translate([-enclosure_height, 0])
                 enclosure_left();
-
+        mirror([1, 0, 0])
             translate([-enclosure_height, enclosure_length + kerf_width])
                 enclosure_right();
-
+        mirror([1, 0, 0])
             translate([-enclosure_height, enclosure_length + kerf_width + enclosure_length_right + kerf_width + enclosure_width - enclosure_horizontal_inset])
                 rotate([0, 0, -90])
                     enclosure_front();
-        }
 
         // Top and bottom
         translate([enclosure_height + kerf_width + enclosure_length_right, enclosure_wall_to_wall_width + kerf_width])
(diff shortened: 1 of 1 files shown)


scottbez1:
This is great, but I think it will need to be an configurable option rather than directly changed.

As I recall, Ponoko cuts MDF with adhesive just on the back side (which gets cut along with the MDF) and then they apply a full adhesive sheet on the top side **after the cut** to hold all the pieces together when removed from the laser cutter and during shipping -- this means that the bottom side avoids burn marks due to the protective film during cutting but the top side does not. The model was originally designed around this, so all "inner" faces are facing up and will get burn marks that can be hidden internally, while the "outer" faces remain clean.

It used to be that Ponoko would recommend placing a note during the order if you wanted the glossy side up while cutting matte acrylic, but it sounds like that's no longer an option? That does have the unfortunate side effect that even if the design is partially mirrored as you've done, it's no longer possible to have the engraving done on the glossy (inner) side of the bottom piece, so it will have to break the consistency of the other faces having the matte side out (or you could put the engraving on the outside).

dmadison:
I didn't realize that Ponoko doesn't use protective paper with MDF. All of my acrylic orders from them have had protective paper on both the top and bottom in addition to the adhesive sheet to keep all of the parts together. That seems like an oversight on their end...

Unfortunately it doesn't look like you can leave an order note or specify the engraving face, likely because orders are now grouped and panelized. Personally I don't mind the etching on the matte side, although if it's bothersome then a user can flip it over, swap the top and bottom pieces, or remove the etch altogether.

Do you suppose the option should mirror the entire 2D design instead of these specific parts? That would make it more obvious which version is which, and would be more future-proof than assuming that the other parts will stay symmetrical.

scottbez1:
I like the idea of mirroring the whole design for simplicity, but won't that mean that any text engravings become unreadable? The challenge is that the text is currently somewhat carefully placed to avoid certain features, so is maybe not so easy to have adapt automatically? (another option is to remove some or all of the text... I added it early on mostly just because I could, though I do like having the commit hash included. Another thing I was thinking about was actually _adding_ text to the parts to help with assembly, since all the inner faces are facing up, at least on the MDF, but of course that gets trickier if matte is always up for the acrylic now...)

So in general I think it would be an acceptable tradeoff to introduce the complexity of selectively mirroring just the front/left/right sides via an option.

dmadison:
All things considered I think adding the option to mirror the whole design is going to be the way to go. It's more future-proof if later portions of the design stop being symmetrical, and it's also simpler to implement since the per-part offsets won't change between the mirrored and non-mirrored versions.

I wouldn't worry too much about the etched text being mirrored. The existing text is easy enough to mirror so it's not backwards, but if other labels start being added they won't work on a mirrored version of the design anyway since they'll be on the outside faces.

I'll submit those changes as a separate PR and close this one, as the differences between the two are significant.

PR #101, opened 2020-11-15
Hall Effect Sensor Spacing and Jig

dmadison:
This PR dynamically adjusts the hall effect sensor's height off of the PCB in the model based on its distance from the spool. The sensor will maintain the same relative distance from the spool as the model is changed and throw an error if its distance from the PCB exceeds the max length of the sensor's leads (per the datasheet). I also created a small lasercut jig that fits between the PCB and the sensor to aid in soldering.

Using your v0.5 sensor assembly video as reference, I set the spool offset to `0.7 mm` which corresponds to a distance of `5.8 mm` from the PCB. Note that this PR has a conflict with #100 that will need to be resolved, as the new jig is cut out of the motor window area.

Here are some demonstration gifs, using a change in material thickness (3.2 -> 1.0) to show the dynamic position.

#### Before:
sf-sensor-position-master

#### After:
sf-sensor-position-pr

And here's the change to the lasercut design ([before](https://user-images.githubusercontent.com/24282108/99187360-39fb2b80-2724-11eb-848e-68b93e99ae2b.png) / [after](https://user-images.githubusercontent.com/24282108/99187365-3b2c5880-2724-11eb-81a0-87e7d7033725.png)).

The spacing jig can be viewed in-place in the 3D design by changing `render_sensor_jig` to 'true'.

### DIFF
--- a/3d/pcb.scad
+++ b/3d/pcb.scad
@@ -16,12 +16,14 @@
 include <m4_dimensions.scad>;
 
 pcb_thickness = 1.6;
+sensor_spool_distance = 0.85;  // distance from the sensor to the face of the spool
 
 // From datasheet:
 hall_effect_height = (2.8 + 3.2) / 2;
 hall_effect_width = (3.9 + 4.3) / 2;
 hall_effect_thickness = (1.40 + 1.60) / 2;
 hall_effect_sensor_offset_y = hall_effect_height - 1.25;
+hall_effect_pin_length_max = 14.5;
 
 // From sensor.kicad_pcb:
 pcb_height = 16.256;
@@ -29,7 +31,6 @@ pcb_length = 16.256;
 pcb_hole_to_sensor_pin_1_x = 8.636;
 pcb_hole_to_sensor_pin_1_y = 1.27;
 sensor_pin_pitch = 1.27;
-sensor_pin_length = 8;
 pcb_hole_to_connector_pin_1_x = 8.636;
 pcb_hole_to_connector_pin_1_y = 8.636;
 connector_pin_pitch = 2.54;
@@ -57,7 +58,7 @@ pcb_sensor_pin_width = 0.43;
 
 
 // 3D PCB module, origin at the center of the mounting hole on the bottom surface of the PCB
-module pcb() {
+module pcb(pcb_to_spool) {
     color([0, 0.5, 0]) {
         linear_extrude(height=pcb_thickness) {
             difference() {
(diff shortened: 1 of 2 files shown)


dmadison:
I was thinking a little more about this and the thickness issue (#93). If your original design was manufactured at 3 mm thick with the CAD thickness set to 3.2, should the spool offset be lowered down to 0.5? That would make the current design offset by 6 mm, which would go down to the accurate 5.8 mm when the thickness is lowered to 3.0.

scottbez1:
> I was thinking a little more about this and the thickness issue (#93). If your original design was manufactured at 3 mm thick with the CAD thickness set to 3.2, should the spool offset be lowered down to 0.5? That would make the current design offset by 6 mm, which would go down to the accurate 5.8 mm when the thickness is lowered to 3.0.

I don't know that it matters enough to be worth preserving that "accurate" 5.8mm value. IIRC the 5.8mm was determined empirically/by eye, and was really specified with more significant figures than relevant, especially considering the leads are being hand-bent and there's typically more than 0.2mm worth of axial slop in the output shaft of the 28byj-48 motors...

dmadison:
That's fair. In that case I'll leave the value as-is. Plus it's probably better to keep the sensor farther away from the spinning spool if possible.

dmadison:
(Closing / reopening to re-trigger the Travis CI build, which apparently failed on the electronics section)

3d/pcb.scad
```diff
@@ -89,11 +97,14 @@ module pcb() {
...
 
-
-
     // Sensor pins
     color([0.5, 0.5, 0.5]) {
-        translate([pcb_hole_to_sensor_pin_1_x - pcb_sensor_pin_width/2, pcb_hole_to_sensor_pin_1_y - pcb_sensor_pin_width/2, -sensor_pin_length + pcb_thickness + 0.1]) {
+        pin_extra_length = 0.1;  // pins excess sticking out from the back of the PCB
+        sensor_z_offset = pcb_to_spool - sensor_spool_distance - hall_effect_thickness/2 - 0.1;
+        sensor_pin_length = sensor_z_offset + pcb_thickness + pin_extra_length;
+        assert(sensor_pin_length < hall_effect_pin_length_max, "Warning: design is too thick to fit sensor");
```
scottbez1, line 105: Oh cool, looks like OpenSCAD finally got bumped to a modern version (2019.05) in Ubuntu 20.04 so we can use `assert` instead of `legacyAssert` now; nice!

scottbez1 (approved):

PR #102, opened 2020-11-24
Etched home flap indicator

dmadison:
As I was starting to put the flaps in my enclosure I thought it would be really nice if there was some sort of indicator for which hole was for the 'home' flap. So I added one:

sf-home-etching-3d

sf-home-etching-2d


scottbez1 (approved):
Great idea!

PR #103, opened 2020-11-29
Flap Case

dmadison:
While I was waiting for the enclosure to be lasercut I got a head start on applying the vinyl to the flaps, but had nowhere to put them. So I designed a 3D printed case that holds a full set of 40 flaps.

sf-flap-case-views

The design includes a finger hole at the bottom for pushing flaps out, a 'pinch' cutout on the sides for grabbing flaps, and a rounded slot on the bottom for retaining a rubber band or other strap to keep the flaps in place. I've printed eight of these (from the Fusion 360 version of the design) and they work great.

The OpenSCAD version submitted is fully modular and will scale in height for different flap counts. I also added an array feature so you can create a single part with multiple rows and columns of containers, such as a 1x4 version for all of the modules on a single driver.


scottbez1 (approved):
Super cool! If you don't mind, want to put together a little bit of content for the README about these 3d-printed jigs? I want to make sure people know about these cool helpers!

PR #104, opened 2020-12-01
Sensor PCB Case

dmadison:
As with the flap case (#103), while I'm waiting for the enclosures to be finished I have a bunch of assembled sensor PCBs sitting around. This is a little dangerous since the hall effect sensors stick out from the board and can be accidentally bent out of position. So I designed a 3D printed case for them to safely sit in until the modules can be assembled.

sf-sensor-case-x1

The design has a `cases_per_row` parameter for generating multiple cases in one design, as well as a `dual_rows` parameter that mirrors a second row along the back for double the number of PCB spots:

sf-sensor-case-x8

### DIFF
--- a/3d/pcb.scad
+++ b/3d/pcb.scad
@@ -56,16 +56,22 @@ pcb_connector_pin_tail_length = 3.05 + 2.5/2;
 pcb_sensor_pin_width = 0.43;
 
 
+module pcb_outline_2d(hole=true) {
+    difference() {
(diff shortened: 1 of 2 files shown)


dmadison:
Printed and with PCBs in place:

sf-sensor-case-x8-printed

scottbez1 (approved):
Wow, super cool! And including fillets in OpenSCAD is some serious dedication!

PR #105, opened 2020-12-04
CAD Render Colors

dmadison:
I've made a number of changes to the assembly to make it easy to adjust the colors.

First, the assembly color variations are now generated from a single base color. This means you can change the one 'assembly' color and still keep the color variation between parts. The scalars were calculated from the current values so everything should keep the same appearance as before. The etch color, similarly, has been broken out into a variable and can now be changed.

Second, I've added variables for the hardware (nuts + bolts) and flap color. This allows you to change out the steel hardware for black oxide or swap the flaps from white to black. The letter color is calculated as the inverse of the flap color, so white flaps get black letters and black flaps get white letters (and red flaps get cyan letters and so on and so forth).

There should be no functional changes to the design.

#### Default: MDF, white flaps, steel hardware

sf-colors-default

#### Demo: Matte black acrylic, black flaps, black oxide hardware

sf-colors-matte-black

Using:
```cpp
assembly_color = [0.1, 0.1, 0.1];
etch_color = [0.5, 0.5, 0.5];
hardware_color = [0.2, 0.2, 0.2];
flap_color = [0, 0, 0];
```

### DIFF
--- a/3d/rough7380.scad
+++ b/3d/rough7380.scad
@@ -26,27 +26,26 @@ module roughM4_7380(length) {
     sphere_offset = 1.2;  // Just a guess
     sphere_radius = sqrt(sphere_offset*sphere_offset + (head_diameter/2)*(head_diameter/2));
 
-    color([0.75, 0.75, 0.8]) {
-        difference() {
-            union() {
-                translate([0, 0, -eps]) {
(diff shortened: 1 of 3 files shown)


scottbez1 (approved):
Neat!

PR #108, opened 2020-12-12
Organize 3D printed tools

dmadison:
Collects all of the 3D printed tool designs in a subfolder to try and keep the CAD files at least somewhat organized. Also fixes a bug with the optional PCB rendering in the PCB case design and adds a section to the README about these designs (per your comment in #103 :smile:).


scottbez1 (approved):
Awesome, thank you!

PR #109, opened 2020-12-14
Zip Ties for Cable Management

dmadison:
Adds holes for cable management zip ties into the left side of the enclosure: one pair underneath the sensor PCB for wires exiting downwards (PCB behind module) and one pair above the motor for wires exiting upwards (PCB mounted on top of module(s)). The goal is to prevent the wires from bending outwards and getting tangled in the moving spools of adjacent modules.

These are sized to work with your typical 2.5 mm wide mini zip ties.

sf-zip-tie-mounts png

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -265,31 +265,6 @@ echo(flap_notch_height=flap_notch_height);
 echo(pcb_to_sensor=pcb_to_sensor(pcb_to_spool));
 
 
-module rounded_square(size, center=false, r=0.0, $fn=$fn) {
-    width  = size[0] == undef ? size : size[0];  // unpack vector if present
-    height = size[1] == undef ? size : size[1];
-
-    if(r <= 0.0) {
-        square([width, height], center=center);
-    } else {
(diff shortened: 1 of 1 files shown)


scottbez1 (approved):
Neat!

Have you tested that the zip tie's natural bend radius on the inside of the enclosure keeps it clear of the flaps? I'd be a little worried that a zip tie will naturally bow out around the bends and interfere with the side of the flaps given how tight that space is. I'm happy to do a test later today if you'd like; sounds like a great opportunity to try out some smaller endmills on my new CNC machine :smile:

dmadison:
Go for it! Sounds like a good excuse to run a few parts :grin:

I haven't tested it myself but I don't expect interference to be an issue. With wires on the opposite side the back portion of the tie should pull pretty flush to the enclosure.

scottbez1:
"pretty flush to the enclosure" is accurate :smile: 

I'm seeing < 1.5mm sticking out on the back side with some dummy wires zip-tied which is more than flush enough.

PXL_20201224_055402052
PXL_20201224_055422982

PR #111, opened 2020-12-22
Alignment Indicators

dmadison:
Adds alignment indicators to the enclosure to aid in assembly: a triangle on the left side, a circle on the right side, and an 'up' arrow on the front face. The goal is to provide an easy visual cue for how the enclosure pieces fit together - particularly the front piece, which is slightly asymmetric. I opted not to add indicators to the top and bottom pieces because they seemed a little redundant.

The triangle and arrow geometry were added through modules included in a new `shapes.scad` file. This also includes the rounded rectangle module that was submitted separately as a part of PR #109 (Zip Ties).

I also cleaned up the other etched features in the design and removed the 3D difference feature for the etched spool indicators which struck me as excessive. All etched features now go through a single helper function to set the positional offset, height, and color.

#### Ponoko Quotes (3.0 mm MDF, 2020-12-21):

| Branch                 | Variant                          | Commit  | Price  |
|------------------------|----------------------------------|---------|--------|
| `master`               |                                  | 0f6d8b7 | $20.29 |
| `alignment-indicators` | one-ended (submitted)    | 71fcfb8 | $20.85 |
| `alignment-indicators` | two-ended                         | 9e60870 | $24.80 |
| `alignment-indicators` | two-ended , line-etched, no label | 9e60870    | $21.55 |

This submitted version only has the one-ended indicators, which add $0.56 USD to the cost of each enclosure or an increase of 2.7%. In my opinion the symmetry of the two-ended indicators looks much better but is not worth the cost increase.

I'm suspecting the large jump in price comes down to how the laser-etching process works. When I used to operate a laser the software would rasterize the design and then run the head back and forth horizontally at a steady pace over all areas with etched items, firing the laser itself only over etched areas. Because the pair of alignment indicators are lined up horizontally the laser is likely performing a continuous slow travel move over the height of the front panel, which would increase the time to cut the design and explain the jump in cost.

For good measure I also tested the two-ended version without the text label and with line-etching instead of area etching, which would presumably reduce the travel time costs but increase the cost of etching itself. Even so, it would still be cheaper to keep the label and use area engraving with one-ended indicators - which is what is submitted.

#### Photos:

sf-alignment-ind-normal

sf-alignment-ind-right

sf-alignment-ind-left

sf-alignment-ind-flat

### DIFF
--- a/3d/shapes.scad
+++ b/3d/shapes.scad
@@ -39,11 +39,8 @@ module rounded_square(size, center=false, r=0.0, $fn=$fn) {
 }
 
 module triangle(size, center=false) {
-    function tri_height(a, c) = sqrt(c*c - a*a);  // pythagoras
-
     width  = size[0] == undef ? size : size[0];  // unpack vector if present
-    height = size[1] == undef ? tri_height(width/2, width) : size[1];  // calculate height with trig if only width provided
-    angle = atan(height/(width/2));  // angle in left corner, degrees
+    height = size[1] == undef ? size : size[1];
 
     pts = [
         [      0,      0],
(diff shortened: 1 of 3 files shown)


scottbez1 (approved):
Amazing! Love the price analysis you did on this - I'm glad you checked 2-sided vs 1-sided because that's a pretty drastic increase in price for what is a small design change!

PR #113, opened 2020-12-23
Spool Letter Rendering

dmadison:
Over-the-top update: this change renders flap letters on both the front and back of *all* flaps in the spool. The program calculates the position of the front-facing letter in the character list, and then adds each front/back letter by iterating through the looped array starting at that point. If the front-facing letter is not present in the character list, only the front letter is rendered and the rest of the spool letters are skipped. This is evaluated per-module so each spool's flaps will have the correct letters in their correct positions.

The existing `render_letters` value has been refactored as `render_string`. The new `render_letters` value controls visibility of the letters themselves: `0` for no letters, `1` for front letters, `2` for all letters. This matches the behavior of the other visibility control options. By default all letters are rendered.

The character set is the same as the the current firmware (` ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.,\`). With Roboto the comma looks a little off as it's visible when hanging below the front flap, but otherwise everything looks as expected.

This serves little functional purpose, but it adds a nice bit of polish to the design + render. My goal was to export the assembly into Blender with all letters using `generate_stl.py`, which works a treat:

#### OpenSCAD:
sf-spool-letters-scad

#### Blender 2.91 (Cycles):
sf-spool-letters-render

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -1068,7 +1068,9 @@ module split_flap_3d(letter, include_connector) {
                             translate([flap_width, flap_pitch_radius, 0]) {
                                 rotate([flap_rendered_angle, 0, 180]) {
                                     flap();
-                                    if((i == 0 && render_letters == 1) || render_letters == 2)
+                                    if(i == 0 && render_letters >= 1)  // for first flap, do not use character list (always match string)
+                                        flap_letter(letter, 1);  // flap on front (top)
+                                    if(i != 0 && render_letters == 2)
                                         flap_letter(get_flap_character(letter, i), 1);  // flap on front (top)
                                     if(render_letters == 2) {
                                         translate([flap_width, 0, -flap_thickness + eps])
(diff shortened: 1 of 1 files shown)


scottbez1:
Sorry for leaving this review hanging. My gut reaction is that it adds complexity that will need to be maintained as things change, but as you say, serves little functional purpose. I suspect that it is also particularly liable to breaking accidentally in the future, since it is so subtle (I found it tricky to validate that the letters were laid out correctly given how little of each is visible).

To me, the primary goal of the model is to generate the actual 2d outputs, and, secondarily, to visualize the design _to the extent necessary to build/modify the design_. That is, components that do not contribute to the output (such as the motor model, the sensor model, screws, etc) are only implemented to make sure that everything fits together correctly, but smaller details like wires, screw threads, etc, are intentionally omitted to keep things simple.

So my default position would be to decline this feature on those grounds, but I am open to discussing further if you would like to make an argument for its inclusion, keeping that framework in mind.

dmadison:
I completely understand where you're coming from, and I agree that it's an inessential contribution. It's a cosmetic change at heart.

As far as functionality the only practical reason I can see for its inclusion is that it gives you a good idea of how the letters look in context with the rest of the spool, particularly with letters like 'Q' which descend further down the flap. Now that the font rendering is accurate you can preview font layouts in real time, including how full words look and how much 'excess' appears on the lower flaps behind the current one. It's been awhile since I've looked at these changes but I think they should work even with modifications to the spool geometry and flap count. Granted that's not a *compelling* reason but it's the best functional argument I've got.

There is a cosmetic argument to be made though. While the CAD design is used for planning and fabricating the display it's also used for visualization and "advertisement" in the sense that it's hosted in the STL viewer on the repo's website. If the viewer can handle a few extra polygons it would look nice with the rendered letters all around the spool.

And while there's some added complexity I think this feature is more robust than it seems. The letter selection is tightly coupled to the existing flap rendering loop, and assuming no one modifies the recursive searching functions I don't foresee a way for things to go awry. (...He says, fully aware of his hubris :stuck_out_tongue_closed_eyes:.)

scottbez1:
Superseded by #154

PR #114, opened 2020-12-30
CAD: Move flap thickness slop to spool slop

dmadison:
This increase was a mistake. The difference in thickness (0.2 on either side) should have been added to the *spool* slop instead of the flap width slop, as the inner width of the spool (flap_width_slop) was not affected by the thickness parameter, while the outer width of the spool (spool_width_slop) *was*. This commit corrects that oversight.

This reverts commit a68fc988bc73c146c28ed1dc2c8e02c990d5ab93 (squashed as part of 7be69dbeb4fb7c1aa4b29027eb3e49a9b121c6a2).

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -117,7 +117,7 @@ letter_color = color_invert(flap_color);  // inverse of the flap color, for cont
 flap_rendered_angle = 90;
 
 
-flap_width_slop = 0.1;  // amount of slop of the flap side to side between the 2 spools
+flap_width_slop = 0.15;  // amount of slop of the flap side to side between the 2 spools
 
 spool_width_slop = 1.4;  // amount of slop for the spool assembly side-to-side inside the enclosure
 
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -117,9 +117,9 @@ letter_color = color_invert(flap_color);  // inverse of the flap color, for cont
 flap_rendered_angle = 90;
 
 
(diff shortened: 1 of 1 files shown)


dmadison:
On my test display the spool will sometimes stall while rapiding. It appears that this is caused by imperfect assembly of the spool and the flaps sliding sideways, causing them to get caught on the upper corner of the front panel. I've traced the issue back to this change and I'm hopeful this PR will fix the issue.

scottbez1 (approved):

scottbez1:
Thanks for following up on this @dmadison and good catch! Do you have any other planned tweaks or changes? If not, I'll go ahead and order a test of the latest 3d designs soon to validate for a new release.

dmadison:
I should have at least one more tweak to make. My own enclosures just arrived so I'll try to unpack and test them this weekend.

dmadison:
Updating for posterity: the original change was *not* a mistake. The `flap_width_slop` value was increased because the *flap dimensions* (notch depth) changed to reflect the die-cut flaps - not because of the thickness decrease. I was confused because the two values should ideally be the same and both are used in the spool width calculation. Had I actually re-read the commit message I referenced (a68fc98) I would have realized my mistake. :flushed:

Nevertheless! Although the comment and reasoning behind the change are wrong the change itself is still correct, as the spool is now tighter on the flaps and should prevent them from sliding out of position and causing the motor to stall. This matches the original design intent of having 0.1 mm (now 0.15 mm) clearance.

PR #116, opened 2021-01-05
fix a problem in the svg duplicate line culling

### LINKED ISSUE #115: The duplicate svg line removal code has a problem
The code that removes duplicate lines from svg paths in `3d/scripts/svg_processor.py` has a problem. When I run `generate_2d.py`, the output in `combined.svg` looks like this:

Screenshot from 2021-01-05 01-42-08

The diagonal lines cutting across some of the parts are caused by trimming lines immediately preceding `Close` elements from some of the paths. When the svg.path package parses an svg path and returns a representation, `Close` elements are returned with a start and end point, just like a `Line` element. But the XML representation of `Close` is just a `Z`, which is interpreted by svg editors as, 'draw a line from the current point to the location of the most recent `Move` element in this path.' As a result, even though the path removal code was updating the start or end points for the `Close` element after it removed the preceding line, those points were just being ignored when svg.path regenerated the XML for the altered path.

I have a fix for this problem, which is just to turn all `Close` elements into `Line` elements when processing paths. The resulting .svg files have worked fine in both svg editors (inkscape and Affinity Design, at least), and when used to drive my Glowforge. I'll submit a PR with those changes in a few minutes.

trapgate:
When culling duplicate lines in the generated svg files, if the path immediately preceding a `Close` element was pruned, the Close would end up drawing a line from the beginning of the deleted element, rather than the end. This is because even though the svg.path library adds a start and end point to a close instruction, the actual XML representation is just `Z`, with no coordinates, and it draws a line from the end of the previous line, to the location of the most recent `Move`. To fix this, this change replaces all `Close` elements with Lines instead.

Fixes #115

### DIFF
--- a/3d/scripts/requirements.txt
+++ b/3d/scripts/requirements.txt
@@ -1,1 +1,1 @@
-svg.path==2.2
+svg.path==4.0
--- a/3d/scripts/requirements.txt
+++ b/3d/scripts/requirements.txt
@@ -1,1 +1,1 @@
-svg.path==4.0
+svg.path==4.0.2
--- a/3d/scripts/svg_processor.py
+++ b/3d/scripts/svg_processor.py
@@ -18,6 +18,8 @@ from collections import defaultdict
 from svg.path import (
     Path,
     Line,
+    Move,
+    Close,
     parse_path,
 )
 from xml.dom import minidom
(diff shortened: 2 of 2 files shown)


trapgate:
As @dmadison pointed out in #116, this occurs only when using a newer version of the svg.path library than the one specified in requirements.txt. So this PR only makes sense if you want to move to the newer svg.path version.

scottbez1:
Thanks for the PR and tips on svg.path changes. I ended up needing to add some code to insert Moves as well, since these aren't inserted or rendered within the path string even when adjacent Lines end up with non-connecting start/end points, so this was resolved in #244

PR #120, opened 2021-01-31
Adjust kerf value

dmadison:
This reduces the default kerf value from 0.18 (0.2 - 0.02) down to 0.1.

I used this value for my last 2 Ponoko orders with 3 mm acrylic and most dimensions are accurate to within +/- 0.06 mm. This has not been tested with Ponoko's MDF.

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -70,9 +70,9 @@ spool_horizontal_explosion = lookup(spool_explosion, [
 ]);
 
 
-// Ponoko kerf values are 0.2 mm for MDF and acrylic (all thicknesses)
+// Kerf adjustment value for 2D export, should be equal to the laser's kerf
 // Remember: it's better to underestimate (looser fit) than overestimate (no fit)
-kerf_width = 0.2 - 0.02;
+kerf_width = 0.1;
 
 // MDF, .120in nominal
 // https://www.ponoko.com/materials/mdf-fiberboard


scottbez1:
I think I'd like to keep the defaults set for MDF for now, since that is the default rendering and the recommended material, given the brittleness of acrylic makes it much less forgiving of tolerance or assembly errors.

I _do_ like having calibrated kerf value presets checked in, but I think having those in render_2d.py as part of the export process would make more sense than as part of the model. Then it would just be a matter of passing an arg like `--kerf-preset ponoko-mdf-3mm` or `--kerf-preset glowforge-mdf-3mm` when preparing for laser cutting. I can get around to that change this weekend.

dmadison:
Although it's untested I'd imagine that the kerf between acrylic and MDF at the same thickness would be similar, at least if Ponoko's specs are to be trusted. They give the kerf as '0.2' for both in the material info page.

I agree that it makes more sense for those settings to be offloaded to the generation script. Part of me wonders if the spool's tab clearance should be specified there as well, given how finnicky that is with acrylic.

scottbez1:
In my experience the kerf values do not match their published specs all that well, and it's always been smaller for acrylic than MDF. I've been using a smaller value of 0.08 for acrylic published files for a while: https://paper.dropbox.com/doc/Ordering-Splitflap-v0.5v0.6--BF_Hq8Ry6FCm9XafR33kAX~MAg-iCwwMo0hRkkAlwdutccWf#:uid=205946111990973782113383&h2=Ponoko

The MDF default value has worked ok on previous orders before the tolerance changes (with some amount of variance; sometimes the spool struts are loose, sometimes they are tight; not all spool struts on the same order fit the same, depending on where there were on the panelized order...), so I'd expect that now that tolerances are included in the design the MDF value may need to be increased ever so slightly (probably about 0.02mm judging by the delta from pre-toleranced 0.08 to your calibrated 0.1 value after tolerances were added).

Regarding the spool tab width: my inclination is to make this a loose fit by default. It's not as nice if you have to glue it, but MDF can be easily glued with elmer's glue or wood glue. And I'm not convinced the consistency of kerf is good enough to get a consistently good interference fit on acrylic (judging by different honeycomb patterns and cut quality I've received on orders from Ponoko I think they may have several different machines with different specs on their production line), so I'd much prefer a loose fit (which should be easily glued with household superglue/CA glue) than one that risks cracking.

dmadison:
Interesting. I haven't ordered my own MDF modules so I'll defer to your experience on that.

And I'll agree about the interference fit on the spool tabs and the acrylic tolerances. On the modules I just received (circa #122) the tabs varied between slightly loose and slightly snug depending on their position in the panel, so I think the design is in a good place right now.

I'll close this PR, pending those "kerf preset" changes to the `generate_2d` script.

scottbez1:
Added in 014be5fd1b24656f460ae1a088b3ef8362257361

PR #121, opened 2021-01-31
CAD: Increase Motor Clearance

dmadison:
Increases the clearance between the motor chassis and the right outside wall of the previous module. Also refactors this variable from 'slop' to 'clearance' for a better description of what it represents.

This is still playing a bit of catch-up from the thickness change (#93). In the previous release where the thickness parameter was set to 3.2 mm but the manufactured panels were actually 3.0 mm thick, there was extra space between the motor chassis and the previous module. This change adds that additional 0.4 mm of clearance back as part of the motor clearance variable.

The additional clearance also makes the widths between the two designs approximately equal - `enclosure_width` is 82.2 mm on v0.6, and 82.25 mm here. (The discrepancy is the additional 0.05 of flap width slop that was added in #114.) Note that #122 bumps this further to 82.35.

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -164,7 +164,8 @@ spool_strut_exclusion_radius = sqrt((spool_strut_tab_outset+thickness/2)*(spool_
 magnet_hole_radius = (4 - 0.1)/2;
 magnet_hole_offset = (spool_strut_exclusion_radius + flap_pitch_radius)/2;
 
-28byj48_chassis_height_slop = 1;
+// Clearance between the motor chassis and the outside right wall of the previous module
+28byj48_chassis_height_clearance = 1;
 
 motor_shaft_under_radius = 0.08;  // interference fit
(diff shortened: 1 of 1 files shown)


scottbez1:
Sorry for taking a while to review these.

I'm confused by the extra 0.4mm though. I can understand adding 0.2mm, due to the adjacent module's wall being that much smaller, but unless I'm mistaken, the thickness shouldn't have any effect on the motor's position, since it's mounted to the _inside_ of the wall, which is the side that's bounded by the top/bottom thickness-independent cuts.

dmadison:
It's because the `thickness` parameter affects the overall enclosure width. By reducing the thickness from 3.2 to 3.0 it reduced the total width of the front panel by 0.4 mm. So although the motor is mounted on the inside face, on the old design the front panel was 0.2 mm wider on that side as well.

scottbez1:
Got it!

PR #122, opened 2021-01-31
CAD: Tweak Spool Fitment Values

dmadison:
On my latest order from Ponoko (3 mm acrylic) the spool struts are still loose and will need to be glued. This increases the interference a bit more (+ 0.04 mm) to hopefully reach a point where the struts will fit into the spool and hold there by friction alone.

Also increased is the flap width slop, in a reversion to the change from #114. On my order based around f4e7714 the flaps do *fit* and swing but it's tight - a little bit of misalignment with the spool end pieces or burrs on the parts will cause them to bind. And after gluing the end pieces in place not *perfectly* parallel (+/- 0.15 mm across ~48 mm diameter) that's exactly what has happened: some flaps on the spool do not 'flip' freely which breaks the whole display.

Originally I thought a slight increase to the flap slop would do the trick, but after giving it some more thought I think the slop decrease should be reverted in its entirety. It's one of the only critical dimensions in the design that can break the entire display if it's too tight, and there are too many factors that can influence the total clearance. Better to play things safe and keep this loose. If there is continued binding due to misalignment of flaps within the spool, the `spool_width_slop` value can be increased further.

I also parameterized and slightly reduced the magnet hole interference. On my displays the magnet press-fit is tight and I ended up cracking one of the spools.  The magnets seemed much happier and still held tight after some light sanding of the hole, so this value is being reduced ever-so-slightly.

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -117,7 +117,7 @@ letter_color = color_invert(flap_color);  // inverse of the flap color, for cont
 flap_rendered_angle = 90;
 
 
-flap_width_slop = 0.15;  // amount of slop of the flap side to side between the 2 spools
+flap_width_slop = 0.25;  // amount of slop of the flap side to side between the 2 spools
 
 spool_width_slop = 1.4;  // amount of slop for the spool assembly side-to-side inside the enclosure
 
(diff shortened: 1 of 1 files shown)


scottbez1:
Good catch; agreed on reverting the flap slop.

PR #124, opened 2021-02-15
Add more tolerance to cavity_top_allowance

lastcoolnameleft:
I think the default should add a bit more tolerance to the hight of the cavity.  I just printed it and it's about 3 cards too low for me.  Perhaps the vinyl added unexpected thickness.  I think 3mm extra should be good.

### DIFF
--- a/3d/tools/flap_container.scad
+++ b/3d/tools/flap_container.scad
@@ -21,7 +21,7 @@ num_flaps = 40;
 containers_x = 1;
 containers_y = 1;
 
-flap_thickness_allowance = 1.1 - flap_thickness;  // fudged, so all flaps are considered 1 mm thick
+flap_thickness_allowance = 1.1 - flap_thickness;  // fudged, so all flaps are considered 1.1 mm thick
 cavity_top_allowance = 2;  // extra case height above the flaps
 
 flap_clearance = 0.5;  // distance between the flap edges and the case walls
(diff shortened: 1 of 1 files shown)


dmadison:
Are you sure you have the front piece on the right way around? It's not symmetrical.

dmadison:
Do you have a pair of calipers to measure the thickness of your vinyl? If the cavity is too short I don't think the right solution is to adjust the amount of extra at the top, it's probably better to change the flap thickness allowance so it works for any number of flaps.

lastcoolnameleft:
The cavity is short by only 3 cards, so I think I have it on the right way around.

Thanks for the suggestion.  I've changed the code so that it changes flap_thickness_allowance instead.  My cards vary between 1.1 and 1.2 depending on the placement and lip of the vinyl.  I'm using cards purchase from @scottbez1 and vinyl stickers from the Ordering guide (Amazon link).

dmadison:
>The cavity is short by only 3 cards, so I think I have it on the right way around.

Yeah sorry about that - in my haste I thought you were talking about the lip on the enclosure. Serves me right for replying without checking what file you modified :sweat_smile:

>I'm using cards purchase from @scottbez1 and vinyl stickers from the Ordering guide

Then I think modifying the thickness is definitely the way to go. I designed the container based off the thickness of my own flaps, and I cut my own letters from a vinyl roll so they're probably on the thinner side.

scottbez1 (approved):
lgtm, thanks!

PR #125, opened 2021-02-20
ESP32 MQTT/WiFi support

johnboiles:
This PR adds support in the ESP32 firmware for connecting to an MQTT server over WiFi for remote controlling the splitflap.

This allows you to do cool things like automate your splitflap using [Home Assistant](https://www.home-assistant.io/).


arduino/splitflap/platformio.ini
```diff
@@ -82,9 +82,11 @@ src_filter = +<*> -<.git/> -<.svn/> -<example/> -<examples/> -<test/> -<tests/>
 
 lib_deps =
     TFT_eSPI@2.2.23
+    PubSubClient@2.8
     ${extra.lib_deps}
 
 build_flags =
+    -DMQTT=true
```
johnboiles, line 89: You might not want this depending on whether you think MQTT or the display will be the primary use case for the ESP32 firmware.

arduino/splitflap/esp32/splitflap_task.cpp
```diff
@@ -200,6 +204,13 @@ void SplitflapTask::runUpdate() {
             break;
         }
       }
+
+      if (recv_buffer_ready) {
```
johnboiles, line 208: There's probably a better way to unify the serial parsing and mqtt message code, but this was quick and easy.

arduino/splitflap/esp32/splitflap_task.cpp
```diff
@@ -265,6 +276,15 @@ SplitflapState SplitflapTask::getState() {
     return state_cache_;
 }
 
+void SplitflapTask::writeBuffer(const char *word) {
+    SemaphoreGuard lock(semaphore_);
+    recv_count = NUM_MODULES;
+    for (int i = 0; i < NUM_MODULES; i++) {
+        recv_buffer[i] = word[i];
+    }
```
scottbez1, line 284: Can you pass the length as an argument and use that to set `recv_count` and the loop to avoid accidentally reading invalid data if an MQTT message doesn't include enough characters?

scottbez1 (commented):
Cool stuff, thanks for the PR! A few questions inline

arduino/splitflap/esp32/main.cpp
```diff
@@ -19,19 +19,29 @@
...
 
 #include "config.h"
-
-#include "display_task.h"
 #include "splitflap_task.h"
 
 SplitflapTask splitflapTask(1);
-DisplayTask displayTask(splitflapTask, 0);
 
+// Note: Currently, MQTT is supported OR a display, not both.
```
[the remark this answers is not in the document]
johnboiles: I don't have one to test with! I'm not that familiar with the FreeRTOS style tasks; if you can run both tasks on the same cpu it probably should just work. The mqtt stuff isn't that timing sensitive

arduino/splitflap/esp32/splitflap_task.cpp
johnboiles: That's definitely better. Good catch

arduino/splitflap/esp32/main.cpp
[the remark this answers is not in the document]
scottbez1: :+1: It should be totally fine to run both the display and MQTT tasks; FreeRTOS does preemptive multi-tasking with, I believe, 1ms scheduler intervals by default in the Arduino/PIO configuration.

arduino/splitflap/platformio.ini
scottbez1: I think for now it'd be best to have MQTT default to disabled, only because it requires additional setup to configure at first, which may not be needed for people that are not planning to use MQTT.

I'm planning to design a new central controller PCB soon based around a TTGO T-Display (which is my favorite ESP32 breakout board and includes a small built-in IPS LCD), so my plan is to have a default configuration for that PCB design, but other options easily configurable (MQTT, BLE, adafruit.io, etc).

johnboiles: Oooh i'll have to pick up one of those TTGO T-Display boards, they look really nice for the price.

arduino/splitflap/esp32/main.cpp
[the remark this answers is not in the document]
johnboiles: Tested and mqtt works just fine with the displaytask enabled!

arduino/splitflap/platformio.ini
johnboiles: Made the change to disable MQTT by default

arduino/splitflap/esp32/splitflap_task.cpp
johnboiles: Made this change in the latest commit.

johnboiles:
Also just to share, here's the current iteration of my split flap:
060680DB-0308-40D1-8ECF-D00CBC631068
08C7AE68-2ABE-48FF-AE93-EF04B61257C9

scottbez1:
I merged in the latest master and resolved a few conflicts. I also simplified the splitflap_task API so there's now just a single showString() method which is used both by the mqtt_task and by splitflap_task itself (but eventually that usage will get pulled out to a separate serial task so this is progress toward that goal). I also fixed a bug with my recent Actions changes, where it was trying to upload to S3 even on forks where the credentials aren't available :facepalm: . Tested this with a local MQTT server - works great! My T-Displays are currently tied up in other projects so I didn't validate that the display still works, but there's no reason to believe it wouldn't.

PR #134, opened 2021-03-15
CAD: Increase connector clearance

dmadison:
On my latest build the edges of the enclosure's front panel are not cut perfectly perpendicular, which effectively increases the front panel width. When using connectors this excess width pushes the modules further apart, causing the display to bow backwards.

This increases the connector clearance so the module connections have a little bit of play, in order to compensate for tolerance issues on the front panel.


scottbez1 (approved):
Sounds good, thanks!

PR #135, opened 2021-03-15
Sensor PCB Tweaks

dmadison:
Fixes the sensor PCB `Edge.Cuts` layer continuity and moves the commit silkscreen off of the pads.

These commits were cherry-picked from PR #97


scottbez1:
Awesome thanks so much for cherry-picking!

I still need to set up the CI to automatically post a comment with modified artifacts, but it should be much easier now that it's running with Github Actions so these artifacts are readily available:
panelized_sensor_merged

PR #140, opened 2021-03-25
Fix firmware compilation for Uno

dmadison:
These pin mode lines were removed as part of the chainlink driver PR (#132) and the consolidation to `LATCH_PIN` for the ESP32, which broke compilation for the "classic" driver on the Arduino Uno. Replacing these lines fixes the compilation error.


scottbez1 (approved):

scottbez1:
Thanks!

PR #142, opened 2021-03-27
Enclosure Alignment Bar

dmadison:
Creates a hole in the left enclosure panel that is aligned with the back edge of the right side of the enclosure, to be used with a dowel rod or other round stock as an alignment aid for multiple modules.

The connector pieces are a bit of a pain to use since you have to remove the front panels and deal with the captive nuts that can be hidden underneath a full spool of flaps. This creates an alternate way to align the back edge of the modules without extending the length of the right panel and increasing the size and cost of the laser-cut design. In larger displays it also allows modules to be swapped out without having to undo any of the hardware.

Note that this does *not* prevent the module from twisting forwards about the alignment hole (along Z), and it would need either an accompanying strip on the front or a bracket attached to the right panel as well. For example, a pivoting latch attached to the mounting bolt or a thin tie around the bottom enclosure piece.

#### Photos:

sf-alignment-bar-wide

sf-alignment-bar-close

sf-alignment-bar-side

### DIFF
--- a/3d/shapes.scad
+++ b/3d/shapes.scad
@@ -71,3 +71,14 @@ module arrow(size, aspect=[0.5, 0.3], center=false) {
             triangle(head);
     }
 }
+
+// positioned where origin is at the point to be rounded (first quadrant)
+module fillet_tool(r, overlap=0.01, $fn=$fn) {
+    tool_size = r + overlap;
+    translate([r, r, 0])
+    mirror([1, 1, 0])
+    difference() {
+        square([tool_size, tool_size]);
+        circle(r=r, $fn=$fn);
+    }
+}
--- a/3d/shapes.scad
+++ b/3d/shapes.scad
@@ -76,9 +76,9 @@ module arrow(size, aspect=[0.5, 0.3], center=false) {
 module fillet_tool(r, overlap=0.01, $fn=$fn) {
     tool_size = r + overlap;
     translate([r, r, 0])
-    mirror([1, 1, 0])
-    difference() {
-        square([tool_size, tool_size]);
-        circle(r=r, $fn=$fn);
-    }
+        mirror([1, 1, 0])
+            difference() {
+                square([tool_size, tool_size]);
+                circle(r=r, $fn=$fn);
+            }
 }
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -630,10 +630,10 @@ module connector_bracket_side_holes() {
 
 module alignment_bar() {
     color(assembly_color1)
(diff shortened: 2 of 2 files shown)


3d/shapes.scad
```diff
@@ -71,3 +71,14 @@ module arrow(size, aspect=[0.5, 0.3], center=false) {
...
+
+// positioned where origin is at the point to be rounded (first quadrant)
+module fillet_tool(r, overlap=0.01, $fn=$fn) {
+    tool_size = r + overlap;
+    translate([r, r, 0])
+    mirror([1, 1, 0])
+    difference() {
+        square([tool_size, tool_size]);
+        circle(r=r, $fn=$fn);
+    }
```
scottbez1, line 83: nit: can you indent all nested operations (or use curly braces), here and elsewhere in this diff

scottbez1 (approved):
Oh, that's really cool!

I think this will also potentially be useful for wire routing if not used for alignment (on that note, I've been thinking about wire routing a lot lately and would love to discuss that further and continue our discussion around potentially flipping the motor and/or redesigning the sensor PCB, though I think Slack will work better than Github for that).

I'm going to prioritize merging #139 ahead of this though, since I think it will be cleaner to resolve conflicts on this smaller self-contained diff than within the large refactor.

scottbez1:
Related to alignment, I should mention that I've been wanting to build a way to optionally export a "merged" front panel as a single piece (and omit the front face as part of the rest of the 2d exports). So you could e.g. order a single 4-module front panel which all 4 enclosure assemblies would share. Is that something that you would find useful?

This is the approach I'm eventually planning to take for the large display, where a front panel will be something 2'x4' (maybe 1/4" MDF for additional stiffness, though may also be screwed into a plywood sub-frame for structural support) and hold a whole bunch of modules, which can be taken out individually for repair if needed by unscrewing the two front screws and pulling the rest of the enclosure assembly (top/bottom/left/right) out from the back.

3d/shapes.scad
dmadison: Done 👍

dmadison:
>I'm going to prioritize merging #139 ahead of this though, since I think it will be cleaner to resolve conflicts on this smaller self-contained diff than within the large refactor.

Sounds great, I'll wait until that's reviewed and merged and then reconfigure this.

> I should mention that I've been wanting to build a way to optionally export a "merged" front panel as a single piece

Great minds... I was thinking the same thing, but having the front panel as an optional 'tool' rather than being integrated into the main module export process. I'm just not quite sure how we would organize the multi-module front panel with the rest of the parts.

scottbez1:
Going to merge now since I think #139 may take longer.

PR #143, opened 2021-03-27
Additional Python Software Scripts

dmadison:
Adds two new Python software scripts:
* **flap_test.py**, which homes the display and then runs through all flaps one at a time (1s between).
* **terminal.py**, which passes strings from input prompts to the display, checking for lower/uppercase alternatives and padding with spaces.

This also moves the `print_status` and `ask_for_serial_port` functions to the `splitflap` module to share them between scripts, and adds a `get_num_modules` function.

### DIFF
--- a/software/demo.py
+++ b/software/demo.py
@@ -1,13 +1,10 @@
 from __future__ import print_function
 
 import random
-import six
 import time
 
-import serial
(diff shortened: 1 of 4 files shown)


scottbez1 (approved):
Thanks!

PR #144, opened 2021-04-10
Add corner relief to punch jig

chrisys:
In order to cater for flaps that aren't cut precisely on the corners and for printers that create a small radius, cut out the corner of the punch jig to allow flaps to sit tightly into the corner.


scottbez1 (approved):
Great idea, thanks for the PR!

PR #156, opened 2021-06-22
Integrate PlatformIO to CI Workflow

### LINKED ISSUE #141: Integrate PlatformIO to CI Workflow
It would be beneficial to integrate some simple compilation checks for the firmware to the CI workflow, to avoid smaller hiccups like #140.

kitlaan:
Followed the instructions at https://docs.platformio.org/en/latest/integration/ci/github-actions.html to add the GitHub Action.
Since I was unsure which platformio environments are active, I included all of them.

Should close #141.

### DIFF
--- a/.github/workflows/pio.yml
+++ b/.github/workflows/pio.yml
@@ -1,0 +1,44 @@
+name: PlatformIO CI
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  pio-build:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Cache pip
+      uses: actions/cache@v2
+      with:
+        path: ~/.cache/pip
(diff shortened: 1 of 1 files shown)


.github/workflows/pio.yml
```diff
@@ -0,0 +1,44 @@
...
+        pip install --upgrade platformio
+
+    - name: Build Firmware
+      run: |
+        pio run -d ./arduino/splitflap \
+          -e uno-shift-register \
+          -e uno-direct \
+          -e mega-direct \
+          -e esp32 \
+          -e chainlink
```
kitlaan, line 44: Since it's my first time dealing with GH Actions, is it better to split these into individual tasks?

scottbez1: I suspect it will be easier to tell which failed if they're separate tasks, but let's start with a single task as-is for now and see how it goes for a while.

scottbez1 (approved):
Awesome! Thanks so much for adding this!

PR #159, opened 2021-07-03
Center front face, tabs, and bolts relative to the window

scottbez1:
Previously the front face and tabs were all centered between the left and right walls, but this resulted in features looking oddly off-center when viewed from the front (where the walls aren't visible and the only other visible feature is the window).

This changes the tabs and front face positioning so that everything is centered relative to the front window. The tabs are made slightly smaller, and the top/bottom pieces are no longer horizontally symmetric as a result. This is not a breaking change (overall dimensions are all the same), but this could not be placed to the _right_ of older modules since the front face has shifted and the motor sticks out further relative to the edge of that face. 

## Comparison
centered_tabs

### Before
centered_tabs_before2

### After
centered_tabs_after2

### DIFF
--- a/3d/splitflap.scad
+++ b/3d/splitflap.scad
@@ -165,14 +165,13 @@ enclosure_wall_to_wall_width = thickness + spool_width_slop/2 + spool_width_clea
 
 // Width of the front panel
 enclosure_width = enclosure_wall_to_wall_width + 28byj48_chassis_height() + 28byj48_chassis_height_clearance - thickness - 28byj48_mount_bracket_height();
-enclosure_horizontal_inset = (enclosure_width - enclosure_wall_to_wall_width)/2;
 front_window_upper_base = (flap_height - flap_pin_width/2);
 front_window_overhang = 3;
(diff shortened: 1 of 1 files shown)


dmadison:
Can you hold off on merging this for a bit? I have another change that I think conflicts.

scottbez1:
Hmmm, I was hoping to merge it soon, it came up when working on my merged front panel diff; this should make it a lot more intuitive when laying out a grid of modules, I think. I can hold off and continue working on a branch if your change is close - what is the change?

scottbez1:
Actually, I may base my prototype CNC cut merged panel off the older pattern for now, so I can test it with my existing v0.7 module, so I'm not in a huge rush to merge after all.

dmadison:
The change is the addition of some 3D printed mounts for securely attaching the modules to a surface (#160). I'd prefer to have that merged before this change, as it makes the mounts available for those using release v0.7. This change will require some small rewrites to accommodate the altered position of the front panel nut.

PR #160, opened 2021-07-03
Mounting Brackets

dmadison:
Creates `mounting_bracket.scad` which generates geometry for 3D-printable mounting brackets. Two brackets are defined, using the same common base geometry:

* `mounting_bracket_t_slot` for mounting a module to an M4-sized T-slot, using an M4-12 bolt and matching T-nut.
* `mounting_bracket_screw_holes` for mounting a module to a wooden board using flat-head M4 screws. The hole for the enclosure's bolt either needs to be undersized for thread engagement or enlarged for a heat-set insert (user's choice).

The base geometry uses the existing hole in the bottom enclosure piece and juts up against the left and right enclosure sides to prevent the module from twisting.

This also adds three more functions to `shapes.scad`: one to create a 3D rounded fillet on a corner, one to create a 3D rounded fillet on an extruded rectangle, and one to create a cone at a specified angle (needed for the 82° countersunk holes).

### T-Slot Mount
sf-mount-adapter-tslot-top

sf-mount-adapter-tslot-bottom

sf-mount-adapter-tslot-positioned

### Screw Mount

sf-mount-adapter-screws

sf-mount-adapter-screws-positioned

For viewing the mounts in position use this snippet in a new file:

```
include <splitflap.scad>
use <tools/mounting_bracket.scad>

position_mounting_bracket() {
    mounting_bracket_t_slot();
    //mounting_bracket_screw_holes();
}
```
I've printed both of these mounts for testing and they work great.

### DIFF
--- a/3d/shapes.scad
+++ b/3d/shapes.scad
@@ -82,3 +82,86 @@ module fillet_tool(r, overlap=0.01, $fn=$fn) {
                 circle(r=r, $fn=$fn);
             }
 }
+
+module fillet_tool_3d(r, r_corner=undef, overlap=0.01, $fn=$fn) {
+    corner_radius = (r_corner == undef) ? r : r_corner;  // use fillet radius if corner radius not provided
+    translate([corner_radius, corner_radius, 0])
+        rotate([0, 180, 90])
+            rotate_extrude(angle=90, convexity=10, $fn=$fn)
+                translate([corner_radius, 0, 0])
+                    rotate([0, 0, 90])
+                        fillet_tool(r, $fn);
+}
+
+module square_fillet_3d(size, r, r_corner=0.0, center=false, $fn=$fn) {
+    width  = size[0] == undef ? size : size[0];  // unpack vector if present
+    length = size[1] == undef ? size : size[1];
+
+    max_radius = min(width, length)/2;  // largest radius possible along XY
+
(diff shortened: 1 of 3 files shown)


scottbez1 (approved):
These look awesome! One comment inline but I'll probably just take care of it myself after merging since it's slightly outside the scope of this PR.

I'm curious - what is the intended way to attach the screw bracket to the module? Would you still use a bolt through the module's mounting hole (along with the 2 other screws on the bracket) into the substrate? Wondering if it makes sense to include a 3d-printed latching post (e.g. something like https://www.thingiverse.com/thing:4796603) instead of a hole? There should be plenty of vertical clearance between the bottom face and the flaps for something like that.

Also, it'd be great to start getting things like this into CI to make sure they don't accidentally break as the main model changes - I created #161 to track.

dmadison:
>I'm curious - what is the intended way to attach the screw bracket to the module?

There's not really enough space below the module for a captive nut, so I thought the user would either undersize the hole and screw directly into the plastic (easy + free but not strong), or oversize the hole and use a heat-set insert (more difficult + costs money, but very strong).

The latching post idea is clever. I wonder how well it would hold up in PLA since the stressed portion would need to be printed in plane with the layer direction.

PR #165, opened 2021-07-07
Tool/model for laying out a merged front panel

scottbez1:
`combined_front_panel.scad` is a model for a configurable merged front panel. The frame width/height are configurable, as are the number of rows & columns of modules and the spacing between modules (can be defined as a "gap" value beyond the minimum possible module spacing, or a center-to-center distance).

There is a matching python script `generate_combined_front_panel.py` for exporting the design as an SVG, for use with a laser cutter or CAM tool.

Also adds some minimal support for routers/endmills with non-trivial kerf - only for the tabs on the front panel - which adds the option of dog bones for the corners and allows for using a tool that is larger than `thickness` while still keeping the vertical mounting position constrained.

Example model rendering for a large (4'x8') layout with 108 modules:
combined_front_panel_example
combined_front_panel_example_closeup

### DIFF
--- a/3d/combined_front_panel.scad
+++ b/3d/combined_front_panel.scad
@@ -102,10 +102,10 @@ for (i=[0:cols-1]) {
     for (j=[0:rows-1]) {
         index = i + j*cols;
         translate([i * layout_center_center_x, -j * layout_center_center_y]) {
-            translate([-flap_width/2, get_flap_gap()/2]) {
+            translate([-flap_width/2, 0]) {
                 flap_with_letters(text_color, flap_color, get_flap_index_for_letter(display_text[j][i]), get_flap_gap());
             }
-            translate([-flap_width/2, -get_flap_gap()/2]) {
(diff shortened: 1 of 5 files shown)


dmadison:
Wow. That's a heck of a panel.

Re: the dog bones, shouldn't the corners of the front window also have those corner drills?

scottbez1:
The front window doesn't need dog bones since nothing square needs to fit into the corners like for the tab slots.

In my infinite wisdom (or more accurately, because I was copying the CR80 card spec) the flaps were designed with ~3mm radius corners on them, so unless you're using a pretty big tool, the curved corners will function just fine. I'm using a 1/8" diameter end mill, so there's plenty of corner clearance. A 1/4" tool would _probably_ be ok too, but maybe pushing it, though you do need to use a smaller tool at some point for the M4 holes regardless, at which point you might as well use that tool to clean up the window corners too.

PXL_20210708_190107100

dmadison:
Ah I see. Somehow I had forgotten about the rounded corners on the flaps.

PR #166, opened 2021-07-11
Classic Driver PCB Mount

dmadison:
This is a 3D printable mounting bracket for the "classic" driver PCB. It includes three mounting points for the PCB itself using heat-set inserts, and two screw mounts on either side for securing into a board.

sf-classic-pcb-mount

In addition to the mount itself this PR makes some quality of life changes to some of the `shapes` functions, including conditional rounding of corners for both the 2D and 3D rounded square functions. I also added the capability to create an "additive" 3D fillet for the rounded square function, seen here connecting the PCB standoffs to the mount base.


scottbez1 (approved):
Cool stuff!

PR #167, opened 2021-07-15
Fix broken sensor PCB holder caused by extraneous PCB

Dinnerbone:
First time using openscad so if this isn't the right way to fix it, please let me know!

The recent motor/sensor testing jig [added an example render of `pcb.scad`](https://github.com/scottbez1/splitflap/pull/157/files#diff-c276f7971839cf96798d592d34f6fb9e3c5dacf62dcc55a6fe809a4b16e78bd9R211) which then snuck into files that were `import`ing `pcb.scad` instead of `use`ing it. This breaks the sensor pcb holder as it tries to render out a dummy pcb in addition to the actual sensor holder.

The options were to either remove that example, or `use` the file. I figured the example was useful if you were to just load up that file in openscad, so I moved out all the dimensions into their own file that can be `include`d, and switched to `use`ing the file.


scottbez1 (approved):
Lgtm! Thanks!

PR #168, opened 2021-07-15
Add $fn parameter to have smoother edges when generating svg fonts

tomihbk:
At the current state, when you use the `generate_font.py` file, the generated svgs are quite low poly. This is due to the facet number on `facet.scad` is set 0 as default. 

If we look at the image below, the first one is what you would get if you generated the fonts. I was asked by the man himself, Scott, to try 60, but it's not quite smooth enough. So I just doubled the value and tried 120. 120 gave me cleaner and more refined results. 

But the [docs ](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Other_Language_Features#.24fa.2C_.24fs_and_.24fn) don't recommend going beyond 100, so this pull request is using 100 but feel free to modify to your needs.

difference_resolution

### DIFF
--- a/3d/flap.scad
+++ b/3d/flap.scad
@@ -23,6 +23,10 @@ use<splitflap.scad>;
 
 character_list = " ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.,'";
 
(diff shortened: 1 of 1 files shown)


scottbez1 (approved):
Looks good, thanks!

PR #171, opened 2021-07-23
Software QOL Changes

dmadison:
Three small quality of life changes to the software script:
* Initializes `character_list` with the default character set, for attempting to filter characters (`in_character_list()`) even if the display is not initalized.
* Adds a `get_text()` helper function, for retrieving the currently displayed string from the status message.
* Adds a serial timeout (1 second), so if the device is not responding the script can be interrupted gracefully without having to kill the process.

### DIFF
--- a/software/demo.py
+++ b/software/demo.py
@@ -21,7 +21,7 @@ def run():
     with splitflap(port) as s:
         while True:
             word = random.choice(words)
-            print('Going to {}'.format(word))
+            print('Going to \'{}\''.format(word))
             status = s.set_text(word)
             s.print_status(status)
             time.sleep(10)
--- a/software/splitflap.py
+++ b/software/splitflap.py
@@ -129,7 +129,7 @@ class Splitflap(object):
 
 @contextmanager
 def splitflap(serial_port):
-    with serial.Serial(serial_port, 38400) as ser:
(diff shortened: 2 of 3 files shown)


scottbez1:
Looks good. I'm curious about the serial interrupt though - are you running this on Windows by any chance? On Ubuntu I'm able to interrupt a serial read/readline when I just tested now. It seems like this may be a [bug in pyserial just on Windows](https://github.com/pyserial/pyserial/issues/343)?

I'm not opposed to setting a timeout as a workaround, but it does require more caution when working with the serial functions going forward. It's not an issue right now since it's just doing naive readlines, but the other thing I wanted to mention on this topic is that I'm planning a complete overhaul of the serial protocol (at least for programmatic communication) in the next month or two, so heads up that I will probably not merge any PRs related to the current serial protocol after this one.

The plan is to migrate to packetized protobuf messages (via nanopb) to get typing and a proper schema instead of the mess of hand-crafted json strings... I will most likely retain some form of plaintext protocol (and retain the json parsing in the python scripts for backwards compatibility, at least for a little while), since it's kind of nice for basic testing and debugging. The plaintext protocol will have a special character command to transition into protobuf binary mode.

dmadison:
I was testing and ran into serial interrupt issues on both Windows and Linux (Raspbian). I wonder if it might be driver related?

PR #190, opened 2021-11-09
Program Problem

DJboyy1:
Hey when i want to execute the generate_2d.py for the svg file for the laser cutter but i always get the same errors and dont know what to do to fix it. Can you please help me 

error

### DIFF
--- a/3d/flap.scad
+++ b/3d/flap.scad
@@ -15,13 +15,12 @@
 */
 include<flap_dimensions.scad>;
 include<global_constants.scad>;
-
 use<flap_fonts.scad>;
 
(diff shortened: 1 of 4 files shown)


scottbez1:
Oops, just saw this now, sorry for taking so long to reply. It looks like you are either missing the `git` program (make sure you can open a terminal and run `git` without any issues) or your copy of the splitflap code is not a git repository (e.g. if you downloaded a zip from github instead of cloning via git) - in order to support a variety of features and make debugging/support easier, a git checkout is the only supported way to use the splitflap codebase. Feel free to open a new issue (not a pull request) if you still have any questions, or join the Slack community using the link on the project's README.

PR #208, opened 2022-04-20
Update package.json

bflatau:
I was getting error messages to add @types/node as a dev dependency when running the typescript build. Might be worth adding to package.json file OR could be something I'm doing wrong....


scottbez1:
Closing this as the js packages have been refactored a decent bit in the meantime and I haven't seen this issue recently or in CI

PR #224, opened 2023-10-24
Fix buffer overrun in showString()

kitlaan:
Found a crash by sending a too-long string via MQTT.

I assume the check in `showString()` was just forgotten, rather than intentional.


scottbez1 (approved):
Thanks!

PR #232, opened 2024-04-04
adding some documentation and a python requirements.txt for ease of use

skofgar:
Hi, I was trying to figure out how to use the demo and thought a little bit of additional documentation could be helpful.


shamlian:
I think you should also add `cobs`, `protobuf`, `pyserial`, and `six` to the `requirements.txt` file. I had to also install these to run the demo.

scottbez1 (approved):
Thanks! I'll follow up with the suggested improvements from @shamlian in a separate change

PR #234, opened 2024-04-07
README.md: Fix link to "a sensor PCB holder"

reedy:


scottbez1:
Thanks for sending this PR! Apologies for not reviewing sooner. In the meantime, the sensor pcb case has been deprecated and removed, so I've resolved this by removing the link entirely instead in 40a672a

PR #238, opened 2024-06-17
Repack components to fit within 20x20cm

scauligi:
As promised, here's a PR with the edits I made!
I removed the `spools_too_big` variable as well, which simplifies the logic a little bit.


scottbez1:
Thanks for this and sorry for not following up for so long. Going to take a look soon and hoping to get this included before cutting the v2 tagged release.

scottbez1 (approved):
This is great! All the rendered artifacts that I checked looked correct! Thanks so much for the effort involved in this!

PR #239, opened 2024-06-20
Elecrow v2 export fixes

len0rd:
PDFs generated for elecrow on the latest master currently look like this:
image

Dimensions appear to be placed incorrectly when the svg output is mirrored. I updated `add_dimensions` to account for mirroring when its enabled.

I tested this in a 20.04 docker container setup similarly to the actions environment and the output looks correct. Result is below:

[elecrow_dim_fix.pdf](https://github.com/user-attachments/files/15907775/elecrow_dim_fix.pdf)

When testing locally, I also discovered that the inkscape cli has deprecated `--verbs`  in favor of `--actions` in version 1.2+. This PR adds a check to handle both inkscape < 1.2 and >= 1.2 when generating elecrow pdfs

### DIFF
--- a/3d/scripts/generate_2d.py
+++ b/3d/scripts/generate_2d.py
@@ -166,14 +166,24 @@ if __name__ == '__main__':
         processor.write(elecrow_svg)
 
         logging.info('Resize SVG canvas')
-        subprocess.check_call([
-            app_paths.get('inkscape'),
-            '--verb=FitCanvasToDrawing',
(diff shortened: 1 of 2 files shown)


scottbez1:
Amazing, thanks for catching and fixing this! Serves me right for hastily enabling mirror by default without re-checking the artifacts...

Running CI now and will merge once confirming the output from Actions. Thanks again!

PR #240, opened 2024-08-19
Update mqtt_task.cpp - added Home Assistant discovery schema

Johnr24:
now the split flap will automatically add itself to home assistant as a text entity.


scottbez1 (approved):
Cool, thanks!

PR #242, opened 2024-09-01
Update Python example (and v2 character set) for new features

shamlian:
- Replaced `~` with `@` in default v2 character set to match shipped flaps
- Get character set from splitflap display (if possible, otherwise use previous legacy alphabet)
- Use upper case in demo words to match new character sets

### DIFF
--- a/software/chainlink/splitflap_proto.py
+++ b/software/chainlink/splitflap_proto.py
@@ -42,14 +42,25 @@ class Splitflap(object):
     RETRY_TIMEOUT = 0.25
 
     # TODO: read alphabet from splitflap once this is possible
+
+    # Flap option 1: Legacy Flaps
+    #_DEFAULT_ALPHABET = [
+    #    ' ',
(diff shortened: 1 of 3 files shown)


software/chainlink/splitflap_proto.py
```diff
@@ -41,15 +41,11 @@ class ForceMovement(Enum):
...
-    # TODO: read alphabet from splitflap once this is possible
-    _DEFAULT_ALPHABET = [
-        ' ',
-        'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm',
-        'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
-        '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
-        '.',
-        ',',
-        '\'',
+    _LEGACY_ALPHABET = [
```
shamlian, line 44: I changed this based on the webserial example.

scottbez1 (approved):
Awesome, thanks for the PR!

PR #256, opened 2025-02-15
3d printable flaps

### LINKED ISSUE #255: 3D Printable Flaps
On discord, the discussion of 3d printed flaps surfaced again.
The color exporter worked fine for me using `python3 3d/scripts/colored_stl_exporter.py 3d/font_generator.scad 3d/build/colored` it spits out stls which can be imported into any slicer and easily colored. (and need to cut to fit on print plates)

@scottbez1 wrote on discord:
> Off the top of my head, I seem to recall a mention of subtracting the letters from the rest of the flap geometry, and then having a single layer of full flap between the front/back letters to prevent light bleeding

The flaps are 0.75-0.8mm, which equals out in 5 0.15 or 0.16 layers. Then 2 layer Font (0.3mm) + solid layer (0.15mm) + 2 layer font (0.3mm).

What needs to be Implemented:
- a flag `3d_print_flag` in flap which triggers font subtraction
- the same flag should also change the font height to 0.3

fidoriel:
closes https://github.com/scottbez1/splitflap/issues/255

### DIFF
--- a/3d/flap.scad
+++ b/3d/flap.scad
@@ -17,6 +17,7 @@ include<flap_dimensions.scad>;
 include<global_constants.scad>;
 use<flap_fonts.scad>;
 use<flap_characters.scad>;
+use<splitflap.scad>;
 
 // TODO: extract core flap spool dimensions used for vertical_keepout_size instead of using the full splitflap file
 use<splitflap.scad>;
@@ -44,9 +45,6 @@ generate_second_half = true;
 
 // index number of flap, see flap_characters.scad
 flap_number = 1;
-
-// gap between flaps
-gap = 5;
 
 // generate 3d printable flap for 0.15mm layer height
 print_3d = false;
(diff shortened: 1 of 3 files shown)


fidoriel:
It seems that a diff is not required. All stls can be imported at once and will be interpreted as a multicolor object.

The diff worked — but just for the non character objects. I don't know why. With the current method, I am able to import everything into a slicer and create multicolor prints with solid layer in between, having the slicer interpreting the diff.

How to do it:

1. Edit `font_generator.scad` and set `print_3d` to true.
2. Run `python3 3d/scripts/colored_stl_exporter.py 3d/font_generator.scad 3d/build/colored_flaps`
3. Drop all parts from output dir into slicer. 
grafik confirm multi color object. OrcaSlicer will be analog.

To separate the flaps with correct color information in place:

1. Rotate the stl so it stands on the bed 
grafik
2. Take the height cut tool and cut at 45mm 
grafik
3. repeat until only one row is left
4. Then rotate so they stand on the bed the other side 
grafik
5. Then cut all at a height of 56mm. 

But this is not great because it takes long time to cut if you need a single flap.
Therefor, I refactored the flap script a little and created a script to generate a export per flap.
`colored_flap_exporter.py`
You need to select the color you want the flap to be in, and then give the flap part the color of the part assembly and select the font in the color you want to have the font in. See screenshot.
grafik

3d/flap.scad
```diff
@@ -39,8 +39,18 @@ letter_facet_number = 100;
 vertical_keepout_mode = 1;              // 0=ignore; 1=highlight; 2=cut
 vertical_keepout_size_factor = 1.1;     // Expand calculated keepout region by this factor. 1=no expansion, 1.5=50% expansion, etc
 
+// generate lower half of the flap
+generate_second_half = true;
```
scottbez1, line 43: Is this needed? I noticed it's set to false in the new script, and I didn't quite understand it's usage in this file (since it adds a second `flap_with_letters`, but `flap_with_letters` should already be double-sided, I think?)

scottbez1 (approved):
Nice, thanks so much for digging into this!

PR #257, opened 2025-02-19
fix(ci): bump action versions to fix JS workflow

not-my-profile:
Since 2025-01-30 the JS workflow has been failing with:

    This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`.
    Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

This commit bumps two action versions to repair the workflow.


scottbez1 (approved):
Thanks for sending!

PR #260, opened 2025-04-06
3d printed flap cleanup

RainyDayHiker:
Base of this pull request is the flap3dcleanup branch - those changes had this description from Scott:

- follow typical generate_ patterns for the script
    - make script executable
    - support font specification and custom text
    - don't require specifying input/output directories
    - export multiple flaps in parallel w/ threadpool rendering
- revert changes to flap.scad example rendering; extract the 3dp-specific setup to flap_3dp.scad file

Additional changes are:
- Update the generate_3d_print_flaps.py file to just call openscad directly for each of the needed files (flap and letters) instead of doing a single call and then trying to split the output by color.
- Update flap_3dp.scad:
  - Change the flap STL to subtract out the letters
  - Add support for generating just the flap or just the letters or both at the same time


3d/flap_3dp.scad
```diff
@@ -0,0 +1,40 @@
...
+flap_number = 0;
+flap_gap = get_flap_gap();
+if (generateFullFlap) {
+	// Generate the full flap using the standard function
+    flap_with_letters([0,0,0], [1,1,1], flap_index=flap_number, flap_gap=flap_gap, bleed=0, print_3d=true);
+} else {
+    if (generateText == false) {
+		// Generate just the flap STL with the letters removed
+        difference() {
+            flap_with_letters([0,0,0], [1,1,1], front_letter=false, back_letter=false, flap_index=flap_number, flap_gap=flap_gap, bleed=0, print_3d=false);
```
pauldhawk, line 33: It might be best to save this as three separate files: top_symbol, bottom_symbol, and flap. When I tried assigning colors by file in my slicer, I had to use the color bucket tool for layers with two colors. However, the paint bucket also applied color to the blank layer between the symbols, which made the backside of the flaps visible from the front.

3d/flap.scad
```diff
@@ -39,15 +39,6 @@ letter_facet_number = 100;
...
 vertical_keepout_size_factor = 1.1;     // Expand calculated keepout region by this factor. 1=no expansion, 1.5=50% expansion, etc
 
-// generate lower half of the flap
-generate_second_half = true;
-
-// index number of flap, see flap_characters.scad
-flap_number = 1;
-
-// generate 3d printable flap for 0.15mm layer height
-print_3d = false;
```
pauldhawk, line 49: I think `print_3d` is needed by the `3d/scripts/generate_fonts.py` and `3d/font_generator.scad` scripts

3d/flap.scad
```diff
@@ -39,15 +39,6 @@ letter_facet_number = 100;
 vertical_keepout_mode = 1;              // 0=ignore; 1=highlight; 2=cut
 vertical_keepout_size_factor = 1.1;     // Expand calculated keepout region by this factor. 1=no expansion, 1.5=50% expansion, etc
 
-// generate lower half of the flap
-generate_second_half = true;
-
-// index number of flap, see flap_characters.scad
-flap_number = 1;
```
pauldhawk, line 46: This is nice to have because if you open the file in OpenScad you can change the flaps by changing this parameter

.gitignore
```diff
@@ -22,3 +22,4 @@ sensor.xml
 *.hex
 
 .pio
+.vscode
```
pauldhawk, line 25: Nice. If you get a sec can you add 
```
.nvmrc
.python_version
```

3d/scripts/generate_3d_print_flaps.py
```diff
@@ -0,0 +1,102 @@
...
+        openscad_variables["generateText"] = True
+        openscad.run(
+            os.path.join(source_parts_dir, "flap_3dp.scad"),
+            os.path.join(flap_path, f"letters.stl"),
+            variables=openscad_variables,
+            capture_output=True,
+        )
+
+        # wait for 10 seconds to allow the user to see the output
+        time.sleep(10)
```
pauldhawk, line 94: nit. 10 seconds is a bit much

3d/scripts/generate_3d_print_flaps.py
```diff
@@ -0,0 +1,102 @@
...
+        help="Clip letters that violate vertical keepout zones",
+    )
+    args = parser.parse_args()
+
+    extra_variables = {}
+    if args.font is not None:
+        extra_variables["font_preset"] = args.font
+    if args.text is not None:
+        extra_variables["character_list"] = args.text
+
```
pauldhawk, line 61: nit. use consistent spacing (remove this line break or put line breaks between the other ifs)

pauldhawk (commented):
Overall looks good. I would run on the python scripts in `.github/workflows/3d.yml` and compare outputs to make sure nothing is breaking.

3d/flap.scad
RainyDayHiker: Things seemed to be continuing to work - I think because it ends up defaulted to false.  But I'm going to add this line to font_generator.scad so that it is documented (and changeable).

3d/flap.scad
RainyDayHiker: Functionality is still there - at the end of that file is an "Example" which would be run if the file is opened directly and there is a variable for the flap number to render.

3d/flap_3dp.scad
RainyDayHiker: This is a great idea!

RainyDayHiker: Added support in the next commit - also added flags to the python script to make it easy to change what to generate (default is 3 STLs).

scottbez1 (approved):
This looks great, thanks! I hope you don't mind, I pushed a few minor changes:
* updated openscad.py to have a `ignore_empty_top_level_object` option - when exporting a single side letter stl for a letter that doesn't have any geometry on that half (e.g. the top half of `.` or either half of a space) I was running into `Current top level object is empty.` errors when exporting, so this allows the script to expect/accept those errors
* add leading zeroes to flap numbers in folder/filenames
* add flap numbers to stl filenames (in addition to the existing numbered folders), so that stls are uniquely identifiable even without context of the folder they were in
* added a `--parallelism` CLI arg (and removed the sleep in the script), so you can easily reduce the parallelism to 1 if you want to debug failures (or set the parallelism to something other than the default of your CPU count)

PR #263, opened 2025-05-13
Avoid SPI input glitches from wifi

scottbez1:
# Changes
* Disable WiFi power saving to avoid sensor data glitches
* Also update platformio espressif platform which was very outdated (note: this may result in a flash rewrite that wipes saved module calibration data!)

# Motivation
* Enabling Wifi previously would cause intermittent "unexpected home" events on arbitrary modules (and sometimes on one of the loopback inputs, causing all modules to be disabled)
* This issue appears related to ESP32 errata 3.11 "When certain RTC peripherals are powered on, the inputs of GPIO36 and GPIO39 will be pulled down for approximately 80ns."
  image
  * Credit to this thread - https://github.com/espressif/arduino-esp32/issues/4903#issuecomment-793442683 - for describing a similar issue and providing a workaround

# Testing
* Correlation with glitches on pin 39 (used as MISO for SPI-based shift-register reads) was identified by probing pin 39 and setting up a trigger on "unexpected home" events
  SDS00024
* Glitches were repeatedly seen with a runt trigger on pin 39
* After applying WiFi.setSleep workaround, no more glitches were captured by the runt trigger, and no more "unexpected home" events were seen

### DIFF
--- a/firmware/esp32/splitflap/http_task.cpp
+++ b/firmware/esp32/splitflap/http_task.cpp
@@ -234,6 +234,10 @@ HTTPTask::HTTPTask(SplitflapTask& splitflap_task, DisplayTask& display_task, Log
 
 void HTTPTask::connectWifi() {
     WiFi.begin(WIFI_SSID, WIFI_PASSWORD);
+
+    // Disable WiFi sleep as it causes glitches on pin 39; see https://github.com/espressif/arduino-esp32/issues/4903#issuecomment-793187707
+    WiFi.setSleep(WIFI_PS_NONE);
+    
     char buf[256];
 
     logger_.log("Establishing connection to WiFi..");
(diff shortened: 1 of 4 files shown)


Johnr24:
Just to mention that the wifi sleep code addition for me even when moving the SPI pin off 39 and onto 26 caused the splitflap to do an unusal homing proceedure when it disconnects from my wifi,

 almost as if the code was causing the wifi to sleep rather than not sleep, espeically odd seeing as I use mysplitflap as a clock and I stream the time into it from an MQTT topic, (so it's changing every minute) 
 
 With the trace drilled out and rerouted to the correct pin and the code changed, removing the more recent wifi code I've foud that I don't get interuptions or module(s) rehoming

PR #271, opened 2026-08-13
Fix broken Documentation Index links in Flaps.md

ctbjdm:
All six "<< Back to Documentation Index" links in `docs/Flaps.md` are 404ing.

They point to `../DocumentationIndex.md`, which resolves to the repo root, but the file lives at `docs/DocumentationIndex.md`. Dropping the `../` makes them relative to the `docs/` directory, matching how `ElectronicsGuide.md` and `MotorGuide.md` already link to the index.

No content changes — path only.


scottbez1 (approved):
Thanks!

